PR #102, opened 2020-05-23
Fixed #98 - Reload ignoring cache with Cmd/Ctrl + Shift + r

### LINKED ISSUE #98: Full refresh shortcut
In Chrome I can do a full refresh with CTRL+SHIFT+r
It would very useful to have that available in the app as well, or pls let me know if I'm just not able to find the same functionality. 
Currently I can't test CSS changes because the app just keeps the cached CSS.

manojVivek:


manojVivek:
@all-contributors please add @manojVivek for code

PR #147, opened 2020-05-30
Added shortcuts to help menu

romanakash:


manojVivek:
@romanakash First of all thanks so much for the PR. Just a couple of thoughts:
1. Is it possible for you to convert this into a popup with HTML content? So that we can present the shortcuts in a more intuitive UI. 
2. If the above is not possible, Just formatting the text in the Popup would do too. In such a way that, the shortcuts section is aligned in line. 

Great work otherwise. 👍

romanakash:
Pushed a new commit with a cleaner looking UI. The messageDialogBox api didn't allow for html content so I used a frameless BrowserWindow. The html content is loaded from app/shortcuts.html

manojVivek:
@romanakash This looks excellent! Can you please add an `ok` button to have close the window?

romanakash:
Oh yes completely forgot about that. Pushed a new commit

manojVivek:
Great, merging this in. Thanks so much for the contribution @romanakash 🎉 
Looking forward to more of these. 

Also, would you mind sharing your twitter handle? We would just like to make a shout out there. 🙂

manojVivek:
@all-contributors Please add @romanakash for code.

romanakash:
Thank you. My Twitter is @romanakash27

PR #162, opened 2020-06-04
Bookmarks

### LINKED ISSUE #37: Bookmarks

Grafikart:
Added a bookmarks bar

- New icon in the address bar to toggle the current url to the bookmarks
- Bookmark bar to display bookmarks

 

## Evolutions

I can fit some improvments if you want

- [x] Allow bookmark deletion with a right click [with material menu ?](https://material-ui.com/components/menus/#menus). I don't know if you already have something for right clicks.
- [x] Allow bookmark renaming (at the moment I use getWebsiteName() from screenshot utils)
- [x] Fill the address bar icon if the website is already bookmarked

## Questions

- I don't know how you would handle getting the title from the url (you would do another request and parse the whole page to querySelector the title ?).
- I kept the separation between container and component but I'm not sure what you want to achieve with this (make the component more testable ?).

Fix #37


manojVivek:
@Grafikart This looks great and that too in short turnaround time. 👏 🎉 

### Regarding Evolutions:
We can ignore the `#2` in the list, as it is not a critical problem. `#1` and `#3` will help complete the workflow with bookmarks. 

### Regarding Questions:
> I don't know how you would handle getting the title from the url (you would do another request and parse the whole page to querySelector the title ?).

We can get the title by running the `querySelector` in any one of the devices, but that is not that big of an issue. We can ignore that for now and stick to the domain logic used for screenshots. And improve it at a later point in time.

> I kept the separation between container and component but I'm not sure what you want to achieve with this (make the component more testable ?).

Yes, keeping the components testable was the idea. But we didn't actually keep up with the unit-tests. 🙂

desktop-app/app/actions/browser.js
```diff
@@ -329,6 +329,20 @@ export function goToHomepage() {
   };
 }
 
+export function gotoUrl(url) {
```
manojVivek, line 332: Is it possible we can just re-use the exiting `onAddressChange` action?

desktop-app/app/components/Addressinput/index.js
```diff
@@ -57,6 +58,25 @@ class AddressBar extends React.Component<Props> {
           onChange={e => this.setState({userTypedAddress: e.target.value})}
         />
         <div className={cx(styles.floatingOptionsContainer)}>
+          <div
```
manojVivek, line 61: The icon in the screenshot looks a bit out of place with the others icons nearby. Shall we try reducing the size a bit?

desktop-app/app/actions/browser.js
Grafikart: Yes totally I fixed it with the last commit

desktop-app/app/components/Addressinput/index.js
Grafikart: For the other icon you rely on a custom SVG apparently. I don't know where you took these icons but I would need to know to find a star SVG with the same stroke density.

Grafikart:
I just added the renaming and reorganised the reducer / actions 

Right click menu on the bookmark item

Rename dialog

desktop-app/app/components/Addressinput/index.js
manojVivek: Those were custom icons put together by combining SVG and cross. Let's keep this on hold, I'll try to find one that matches here.

manojVivek:
@Grafikart Looks great, Is it possible to make the option as "Edit" instead of "Rename" and also show the URL in the popup? In case the user wants to edit the URL.

jjavierdguezas:
@Grafikart can you add a shortcut for this action? it could be `mod+d`.. check [here](https://github.com/manojVivek/responsively-app/blob/e3b0f847c2033618dfdc5c4ba7f0dd7f0d6d040b/desktop-app/app/containers/Root.js#L75)

something like: 
```javascript
    registerShortcut({id: 'AddBookmark', title: 'Add Bookmark', accelerators: ['mod+d']}, () => {
      const currentUrl = ... ; // get current url
      store.dispatch(toggleBookmarkUrl(currentUrl));
    }, true);
```
(cc: @manojVivek )

Grafikart:
I'll do this tonight.

Grafikart:
I merged master and added the shortcut and the ability to update the URL.

manojVivek:
@Grafikart Thanks for keeping it up to date. Can it be merged once the last checklist item is done? 

> Fill the address bar icon if the website is already bookmarked

Grafikart:
@manojVivek I changed the size of the bookmark icon and it is now filled when the current page is bookmarked. It's ready to merge .

desktop-app/app/components/Addressinput/index.js
Grafikart: I added small font size so it first a bit more but the stroke is still too wide.

manojVivek:
@Grafikart Thank you so much for all the hard work on this, I'm merging this in now. 🙌

manojVivek:
@all-contributors Please add @Grafikart for code.

PR #166, opened 2020-06-06
Add shortcuts handler infrastructure

jjavierdguezas:
# Overview
I wrote a piece of code that allows us to register shortcuts from the `Renderer` process and store them in the `Main`  process, all using IPC communication, so we would have a clean way of dispatching actions with shortcuts and having the list of shortcuts being used in the app at any time (which can be used in the hotkeys dialog). We can store as much shortcut information as we want to better display them in the hotkey dialog. If we move forward with this idea it could be the basis for the user to define their own shortcuts in the future.

This PR has the minimum amount of code that implements the idea.

# Things to think about
- Right now this is relying on `globalShortcut` from `electron` and because it runs in the `Main` process, the managers communicate using IPC. Maybe we can call `globalShortcut` from `remote` in the `Renderer` process and only notify `Main` in order to store the shortcut infos and let `electron` to handle the shortcut registration via its own IPC channels. I went with the custom option to have more control and make it easier to change the infrastructure in the future.
- Maybe `globalShortcut` is not the best shortcut manager (for example I couldn't get `Ctrl + -` shortcut to work) and we could use another lib in the `Renderer` process (something like [this](https://www.electronjs.org/docs/tutorial/keyboard-shortcuts#shortcuts-within-a-browserwindow))
- Maybe we should put all shortcuts registration in one method and call it just after `initRendererShortcutManager` call
- I don't know if the scroll in the hotkeys dialog looks good

### DIFF
--- a/desktop-app/app/menu.js
+++ b/desktop-app/app/menu.js
@@ -85,7 +85,7 @@ export default class MenuBuilder {
           });
 
           win.on('blur', () => {
-             win.hide();
+            win.hide();
           });
 
           win.on('closed', () => {
--- a/desktop-app/app/menu.js
+++ b/desktop-app/app/menu.js
@@ -10,6 +10,7 @@ import {
 } from 'electron';
 import * as os from 'os';
 import {pkg} from './utils/generalUtils';
+import {getAllShortcuts} from './shotcut-manager/main-shortcut-manager';
 
 const path = require('path');
 
@@ -68,6 +69,7 @@ export default class MenuBuilder {
             webPreferences: {
               devTools: false,
               nodeIntegration: true,
+              additionalArguments: [JSON.stringify(getAllShortcuts())]
             },
           });
 
@@ -83,7 +85,7 @@ export default class MenuBuilder {
           });
 
           win.on('blur', () => {
-            win.hide();
+             win.hide();
           });
 
           win.on('closed', () => {
--- a/desktop-app/app/shotcut-manager/main-shortcut-manager.js
+++ b/desktop-app/app/shotcut-manager/main-shortcut-manager.js
@@ -4,6 +4,7 @@ import {
     UNREGISTER_CHANNEL,
     REGISTER_REPLY_CHANNEL,
     UNREGISTER_REPLY_CHANNEL,
+    GET_ALL_CHANNEL,
     CLEAR_CHANNEL,
     ShortcutDefinition,
     validateDefinition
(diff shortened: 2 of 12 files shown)


manojVivek:
@jjavierdguezas Thank you so much for coming up with the idea for a clearer way to manage shortcuts and also implementing it. 🎉

desktop-app/app/menu.js
```diff
@@ -68,6 +69,7 @@ export default class MenuBuilder {
             webPreferences: {
               devTools: false,
               nodeIntegration: true,
+              additionalArguments: [JSON.stringify(getAllShortcuts())]
             },
```
manojVivek, line 73: @jjavierdguezas Should we also look at registering the `accelerators` defined below in this file via the shortcuts-manager module?

desktop-app/app/shotcut-manager/main-shortcut-manager.js
```diff
@@ -0,0 +1,66 @@
...
+    return validateDefinition(shortcut) && !(checkUnique && shortcuts.has(shortcut.id));
+}
+
+export function registerShortcut(shortcut: ShortcutDefinition): boolean {
+    if (!validate(shortcut, true)) return false;
+    const { id, accelerators } = shortcut;
+    shortcuts.set(id, shortcut);
+    let ok = true;
+    accelerators.forEach(acc => {
+        ok = ok && globalShortcut.register(acc, () => mainWindow.webContents.send(id));
```
manojVivek, line 26: @jjavierdguezas I think we should not use `globalShortcut` for all the accelerators. 

`globalShortcut` will hijack this short cut at the OS level so it might affect other apps that have similar key combinations registered as a shortcut. We ran into a problem when we used `globalShortcut` for `Ctrl+R` for reload. It was preventing the shortcut functionality in Chrome to work when Responsively was open. 

So we fixed it with the `accelerators` field in the `Menu` object. There should we a way to register these as app-level shortcuts instead of global so that it will be invoked only when the app is in focus.

jjavierdguezas: oops, I hadn't thought of that...good catch!
maybe we can use [this](https://github.com/parro-it/electron-localshortcut)

desktop-app/app/menu.js
jjavierdguezas: maybe.. if you look into the hotkeys dialog HTML file I handle those "static" shortcuts separately from the "dynamic" ones. Do you want to store them all in the manager?

jjavierdguezas:
@manojVivek I updated this branch, removing the `globalShortcut`. I wanted to use the `electron-localshortcut` package but it seems that it is outdated (as well as some of its dependencies). So I copied the code of those packages and modified them until I made it work.
I am no longer convinced that this is the easiest way to achieve this 😓. Looking at the resulting code, I think there is too much code to just register shortcuts in the renderer process and store them in the main.
I will think about another approach to implement this to see if I can simplify the development. My idea is to try with a library in the renderer process and just notify the main process (via IPC) of the shortcuts registered to store them

jjavierdguezas:
@manojVivek  since #168 is merged should I close this PR?

manojVivek:
@jjavierdguezas I'll close it. 👍

Issue #90, opened 2020-05-21
Suggestion to add hotkeys


Zerstoren-ak:
Hello there, first of all – thank You for the app, it's so good)

I just wanted to suggest to add those kind's of keyboard hotkeys and shortcuts, which we all are used to, like: Opt + Command + I for "Inspect Element"; "Zoom in/out" with Command + +/-;  etc.

Thank You.

manojVivek:
@Zerstoren93  Great suggestion, will incorporate these in the future release.

Zerstoren93:
Thank's for response)

Also, there is one more cool feature You can add to your app to realise not only devise simulation but cross-browsing as well, may be you have seen it - playwrigth: https://github.com/microsoft/playwright. Check it out, it could be very useful.

drbyte:
Adding `CMD+L` for editing the URL would be useful too.

romanakash:
Can I work on this?

manojVivek:
@romanakash I think @jjavierdguezas is already working on it. You can take a look at #31 if you can. 

Thank you much for checking about it. 🙂

jjavierdguezas:
yes @romanakash  I'm working on this, please @manojVivek  assign me this issue

jjavierdguezas:
@manojVivek  since #171  is merged I think this issue can be closed

manojVivek:
@jjavierdguezas Thanks, closing it. This will go live tonight. 👍

PR #168, opened 2020-06-07
Develop shortcut managers without globalShortcut or localShortcut

jjavierdguezas:
Using mousetrap instead
Automatically register menu items with accelerators
Fix main window losing focus when hotkeys dialog closes
related to #90 
rewrote of #166

### DIFF
--- a/desktop-app/app/containers/Root.js
+++ b/desktop-app/app/containers/Root.js
@@ -11,6 +11,9 @@ import {grey} from '@material-ui/core/colors';
 import {themeColor} from '../constants/colors';
 import ErrorBoundary from '../components/ErrorBoundary';
 
+import {registerShortcut, clearAllShortcuts, unregisterShortcut} from '../shortcut-manager/renderer-shortcut-manager';
+import {onZoomChange} from '../actions/browser';
+
 type Props = {
   store: Store,
   history: {},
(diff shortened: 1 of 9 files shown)


jjavierdguezas:
@manojVivek I completely rewrote this idea without using `globalShortcut` or `localShortcut`.
I used [Mousetrap](https://craig.is/killing/mice) instead and the result is the same functionality but much simpler. I let you decide wich branch to merge, but I recommend this one, as far as I am concern we can drop the other one (#166)

manojVivek:
@jjavierdguezas I'm with this PR. Have you tested that the existing shortcuts are working as it is and this change is not creating any issues shortcuts in other apps when Responsively is open right? 

Just wanted to confirm before merging it. 🙂

jjavierdguezas:
> @jjavierdguezas I'm with this PR. Have you tested that the existing shortcuts are working as it is and this change is not creating any issues shortcuts in other apps when Responsively is open right?
> 
> Just wanted to confirm before merging it. 🙂

@manojVivek yes I tested it all and it does not mess with other apps. Registered shortcuts only work if the ResponsivelyApp main windows is focused. The shortcuts in the menu items still work just fine. In my last commit I fixed the accelerator for the `Open HTML file` menu item that did not work in Windows (it was defined as `Command+O` but `Command` does not work in Windows, I changed it to `CommandOrControl+O`). On the other hand, I think you should review the menu definition, because I saw that it looks different (it shows different menus and actions) when the operating system is darwing compared to windows and linux

manojVivek:
@jjavierdguezas Thanks for taking care of the Open HTML bug too. 
Yeah, `menu.js` needs cleanup, it has different menu items for windows and Mac. But we keep it for later. :) 

I'm merging this PR now.

jjavierdguezas:
> @jjavierdguezas Thanks for taking care of the Open HTML bug too.
> Yeah, `menu.js` needs cleanup, it has different menu items for windows and Mac. But we keep it for later. :)
> 
> I'm merging this PR now.

@manojVivek  great!, then I'll start working on registering some shortcuts like the ones in #90 , you can assign it to me

manojVivek:
@jjavierdguezas I'm not able to assign it since you have not interacted directly with the card. Can you make a comment on it? 

Thanks!

jjavierdguezas:
> @jjavierdguezas I'm not able to assign it since you have not interacted directly with the card. Can you make a comment on it?
> 
> Thanks!

done

PR #174, opened 2020-06-09
fix windows open-url

Nothing-Works:
I believe I have fixed `open-url` feature on `Windows`. I have tested on both `Windows` and `Mac`. 
And this is related https://github.com/manojVivek/responsively-app/issues/155
Fix https://github.com/manojVivek/responsively-app/issues/155
Please let me know if I am missing something or something can be improved.

### DIFF
--- a/desktop-app/app/main.dev.js
+++ b/desktop-app/app/main.dev.js
@@ -90,6 +90,9 @@ const openUrl = url => {
  */
 
 app.on('will-finish-launching', () => {
+  if (process.platform === 'win32') {
+    urlToOpen = process.argv.filter(i => /^responsively/.test(i))[0];
+  }
   if (['win32', 'darwin'].includes(process.platform)) {
     if (process.argv.length >= 2) {
       app.setAsDefaultProtocolClient(protocol, process.execPath, [
@@ -196,8 +199,9 @@ const createWindow = async () => {
     if (urlToOpen) {
       openUrl(urlToOpen);
       urlToOpen = null;
+    } else {
+      mainWindow.show();
     }
-    mainWindow.show();
   });
 
   ipcMain.on('http-auth-promt-response', (event, ...args) => {


desktop-app/app/main.dev.js
```diff
@@ -196,8 +199,9 @@ const createWindow = async () => {
     if (urlToOpen) {
       openUrl(urlToOpen);
       urlToOpen = null;
+    } else {
+      mainWindow.show();
```
Nothing-Works, line 203: The reason why I add `else` is that there is `mainWindow.show();` in `openUrl()` and two `mainWindow.show();` calls is going throw an error on my machine.

desktop-app/app/main.dev.js
```diff
@@ -90,6 +90,9 @@ const openUrl = url => {
  */
 
 app.on('will-finish-launching', () => {
+  if (process.platform === 'win32') {
+    urlToOpen = process.argv.filter(i => /^responsively/.test(i))[0];
+  }
```
Nothing-Works, line 95: here I am checking if it's `Windows`, we are going to assign the `url` to `urlToOpen`.

manojVivek:
@Nothing-Works Thanks for putting in the time and fixing it, really appreciate it! 🎉 

Just one clarification before merging this, now the extension opens the URL from browser both when the app is open and closed, right?

Nothing-Works:
@manojVivek From what I have tested when the app is closed, it's going to open the app with the current browser URL.

If there is an open app:
 - `mac`: it's going to show the current `mainWindow` with correct URL
 - `windows`: it's going to open another electron instance with correct URL. And this was the previous behavior when you open the app multiple times on `Windows`.

If you want we can make it that there will be only one instance on `Windows` as well.

manojVivek:
@Nothing-Works Yeah, I think, it makes more sense to open it in the `mainWindow` for windows as well.

jjavierdguezas:
@manojVivek could the user decide via options?
- Open in current instance (if there is any)
- Open in new instance

manojVivek:
I just that will keep it simpler, UX wise, and didn't want to confuse the user with more options. :) 

The user can anyways click on the `Back` button to go back to the URL that was open before this.

@jjavierdguezas  What do you think?

jjavierdguezas:
> I just thought that will keep it simpler, UX wise, and didn't want to confuse the user with more options. :)
> 
> The user can anyways click on the `Back` button to go back to the URL that was open before this.
> 
> @jjavierdguezas What do you think?

yeah.. simpler is better and what you say makes sense

manojVivek:
@Nothing-Works Thank you so much for staying committed and making the changes. Looking forward to more like this. Merging this in. 🎉

manojVivek:
@all-contributors Please add @Nothing-Works for code.

manojVivek:
@Nothing-Works Would you mind sharing your twitter handle? We would love to send a shout out on twitter for this work! 🙂

Nothing-Works:
@manojVivek Hi I am very sorry that the last push was just a `rebase` as I wanted to get the latest updates from the master. The code was the same as before, which everything works but on `Windows` it is going to have another instance when you run the app. I am very sorry about this, should I have another that make it a single instance on `Windows`?

Nothing-Works:
@manojVivek https://twitter.com/AndySong666 I will be very happy if you can do so, thank you very much.

manojVivek:
> @manojVivek Hi I am very sorry that the last push was just a `rebase` as I wanted to get the latest updates from the master. The code was the same as before, which everything works but on `Windows` it is going to have another instance when you run the app. I am very sorry about this, should I have another pr that make it a single instance on `Windows`?

No worries, it will be great if you could do that. thanks! :)

PR #178, opened 2020-06-09
add check for updates menu item

### LINKED ISSUE #175: update indicator/ check for update
I think we should add for checking new versions just like vscode.

And when it's downloading the new version in the background, we should also show something like vsvode as well.

jjavierdguezas:
resolve #175

### DIFF
--- a/desktop-app/app/menu.js
+++ b/desktop-app/app/menu.js
@@ -11,7 +11,7 @@ import {
 import * as os from 'os';
 import {pkg} from './utils/generalUtils';
 import {getAllShortcuts, registerShortcut} from './shortcut-manager/main-shortcut-manager';
-import {autoUpdater} from 'electron-updater';
+import {appUpdater, AppUpdaterState} from './app-updater';
 
 const path = require('path');
 
@@ -96,8 +96,9 @@ export default class MenuBuilder {
       },
       {
         label: 'Check for Updates...',
+        id: 'CHECK_FOR_UPDATES',
         click() {
-          autoUpdater.checkForUpdatesAndNotify().then(r => {
+          appUpdater.checkForUpdatesAndNotify().then(r => {
             if (r == null || r.updateInfo == null || r.updateInfo.version === pkg.version) {
               dialog
                 .showMessageBox(BrowserWindow.getAllWindows()[0], {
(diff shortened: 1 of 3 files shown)


Nothing-Works:
@jjavierdguezas hi, good job. I did not test it locally, but just one thing that I want to check, when it's downloading the new version in the background can we show something like "updating..." for the menu? or does your feature already doing this?

jjavierdguezas:
@Nothing-Works when there is an update a System notification is displayed with a descriptive message, I think that is enough, but if you have another idea, please share it and I'll update the PR

jjavierdguezas:
@Nothing-Works I think that it is possible to ask the user if he wants to close the app to trigger the update installation..
also I think that subscribing to the updater events we can know in which state the update downloading is, but right now I don't know where to show that information.. do you have any idea? maybe @manojVivek ? maybe we can wait for #153 and put this into the status bar... what do your think?

manojVivek:
@jjavierdguezas I was thinking the same, this will be a great candidate to be shown in the status bar.

desktop-app/app/menu.js
```diff
@@ -93,6 +94,21 @@ export default class MenuBuilder {
           });
         },
       },
+      {
+        label: 'Check for Updates...',
+        click() {
+          autoUpdater.checkForUpdatesAndNotify().then(r => {
```
manojVivek, line 100: @jjavierdguezas `autoUpdater.checkForUpdatesAndNotify()` is already invoked from the `main` process during every load. 

Any idea what this will return if the update download is already completed from that call?

desktop-app/app/menu.js
```diff
@@ -93,6 +94,21 @@ export default class MenuBuilder {
...
+            if (r == null || r.updateInfo == null || r.updateInfo.version === pkg.version) {
```
manojVivek, line 101: Maybe in the `else` we should show an appropriate message saying "Update in progress" or something like that?

desktop-app/app/menu.js
jjavierdguezas: I clicked several times the menu item (during and after the call to `checkForUpdatesAndNotify`  from app load) and always the same behavior: The same object (`r`) with the same info is returned and a notification is displayed (if the app is outdated). I really don't know if it downloads (and overwrite) the update every time or if it does something smart
From the [documentation](https://www.electron.build/auto-update):
>appUpdater.checkForUpdates() ⇒ Promise<UpdateCheckResult>
>Asks the server whether there is an update.
>
>appUpdater.checkForUpdatesAndNotify() ⇒ Promise< | UpdateCheckResult>
>Asks the server whether there is an update, download and notify if update available. 
>
>appUpdater.downloadUpdate(cancellationToken) ⇒ Promise<any>
>Start downloading update manually. You can use this method if autoDownload option is set to false.

Maybe, if we are going with a complete custom solution, we can call `checkForUpdates` instead and then call to `downloadUpdate` to download it manually then restart the app. but I think this is over-complicate a flow that is handled by `electron-updater`... what do you think @manojVivek  ?

desktop-app/app/menu.js
jjavierdguezas: Yes we could, I just followed the `vscode` behavior: if there is no update a dialog is shown and if there is an update a notification is shown. In `vscode` the notification is not a system notification but I though that it would be fine, because the notification shows a very descriptive message. But I'll go with whatever you want @manojVivek  , just tell me your idea

desktop-app/app/menu.js
manojVivek: Yeah, let's not go that far. We should let `electron-updater` do it automatically.

manojVivek:
@jjavierdguezas The menu's not updating while displayed is reasonable. Not a big issue, I guess.

Shall I merge this?

jjavierdguezas:
> @jjavierdguezas The menu's not updating while displayed is reasonable. Not a big issue, I guess.
> 
> Shall I merge this?

yes, this PR is ready to merge

manojVivek:
Thanks, JJ. Merging it. 👍

PR #179, opened 2020-06-09
Fix app automatically navigating to iframe URL

### LINKED ISSUE #164: App automatically navigates to iframe URL if lazy loaded
Here's a URL to test it:
[https://hallettspoint.com](https://hallettspoint.com)

Scroll down and an iframe will spawn in the DOM (lazy loaded)

**Expected behavior:**
The iframe would remain embedded on the parent page.

**Actual behavior:**
Once the iframe source URL loads, it loads in the URL location bar, effectively leaving the page.

I've confirmed that this only seems to happen if an iframe is lazy loaded. I'm not positive if it has something to do with the particular iframe I'm loading. Can anyone else reproduce this?

RomainFrancony:
As said in #164, we just needed to check `isMainFrame` in the `did-navigate-in-page` event handler.

Also some space where removed in the code during the commit, sorry 😬 

fix #164

### DIFF
--- a/desktop-app/app/components/WebView/index.js
+++ b/desktop-app/app/components/WebView/index.js
@@ -141,8 +141,8 @@ class WebView extends Component {
       }
     );
 
-    const urlChangeHandler = ({url}) => {
-      if (url === this.props.browser.address) {
+    const urlChangeHandler = ({url, isMainFrame}) => {
+      if (!isMainFrame || url === this.props.browser.address) {
         return;
       }
       this.props.onAddressChange(url);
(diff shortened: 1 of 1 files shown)


manojVivek:
@RomainFrancony This is really great. 
Yeah, we need to iron out the linting, formatting, and the commit hook configuration. 😅

joaottc:
Running version Version 1.16.0 and I can confirm that this still happens, in my case when using an iframe from tiktok player such has:

```jsx
const getSrc = (id: string) => {
  const src = new URL(`https://www.tiktok.com/player/v1/${id}`);

  Object.entries(options).forEach(([key, value]) => {
    src.searchParams.append(key, value === false ? '0' : '1');
  });

  return src.toString();
};

export const TiktokEmbed = ({ videoId }: Props) => {
  return <iframe height="100%" width="100%" src={getSrc(videoId)} />;
};

```

amenk:
we have the problem, too on 1.16.0. @joaottc I guess we should open a new issue so this does not get unnoticed? 
is there a way to download an older build to verify if it was a regression?

PR #183, opened 2020-06-10
refactor: react tabs using tab panel for individual layout

sprabowo:
Hi @manojVivek,

just want to refactor code for Individual Layout based on react-tabs using Tab Panel, so there's no error in development when switching to Individual layout ([https://github.com/manojVivek/responsively-app/issues/182](https://github.com/manojVivek/responsively-app/issues/182))

Thanks

### DIFF
--- a/desktop-app/app/components/DevicesPreviewer/index.js
+++ b/desktop-app/app/components/DevicesPreviewer/index.js
@@ -35,7 +35,7 @@ export default function DevicesPreviewer(props) {
                   return null;
                 }
                 return (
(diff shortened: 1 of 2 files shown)


sprabowo:
@manojVivek Thanks for giving the more detail and expectation.

i still miss to solve this test case. i attach my screenshot what actually do in this PR. the tabs were not being synced for hamburger menu in google.com. see video [https://youtu.be/mE9dqfT4OhQ](https://youtu.be/mE9dqfT4OhQ)

maybe i close this PR

PR #191, opened 2020-06-14
Fix ugly global scrollbar in hotkeys dialog

jjavierdguezas:
This only happens in the production build 🤷🏻‍♂️

Before:
image

After this fix:
image

- Fix: set `max-height` property to `ul` inside `shortcuts.html`


manojVivek:
Great! 👏

PR #253, opened 2020-07-03
Fix horizontal scrolling

### LINKED ISSUE #233: No horizontal scroll bar when big enough device on individual layout
>Version: 0.5.0
Electron: 8.3.0
Chrome: 80.0.3987.165
Node.js: 12.13.0
V8: 8.0.426.27-electron.0
OS: Windows_NT x64 10.0.18362

When you are in individual layout and the focused device is bigger than the screen `ResponsivelyApp` should show an horizontal scroll bar (at app level)
A workaround right now is to use `Shift + mouse wheel` to scroll (at least on Windows)

scroll bar issue

RomainFrancony:
Prevent the main content div from growing more than its container and thus hiding the scrollbar behind the status bar.

I also made the devices tabs scrollable, they were shrinking and became ugly on small screen.

fix #233

### DIFF
--- a/desktop-app/app/app.global.css
+++ b/desktop-app/app/app.global.css
@@ -79,13 +79,26 @@ a:hover {
   cursor: pointer;
 }
 
+.react-tabs {
+  position: sticky;
+  left: 0;
+  top: 0;
+  margin-bottom: 10px;
+  z-index: 4;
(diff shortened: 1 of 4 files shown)


manojVivek:
Great, thanks for the quick fix @RomainFrancony!

Issue #237, opened 2020-07-02
Add support for devtools extension that are not released in the Chrome web store


prashantpalikhe:
Often, we have custom devtools extensions for our own projects that for obvious reasons, we won't publish to the web store.

It would be great to add support for custom extensions like that. Just like how Chrome allows us to add custom extensions in developer mode.

jzabala:
Closed by #259

PR #259, opened 2020-07-03
feat: add support for local devtools extensions

prashantpalikhe:
Implement #237


prashantpalikhe:
Decided to use the [BrowserWindow.addDevToolsExtension](electron-devtools-installer) to keep the changes minimal. Also, the version of [electron-devtools-installer](https://github.com/MarshallOfSound/electron-devtools-installer/blob/v3.0.0/src/index.js) that's being used is using the same under the hood. 

Moving to `session.loadExtension` can be a part of another migration dedicated to that. We need to work with `extensionId` then, which we don't atm. 

WDYT?

desktop-app/app/components/ExtensionsManager/index.js
```diff
@@ -58,13 +59,19 @@ export default function ExtensionsManager({triggerNavigationReload}) {
     setLoading(true);
     setErrorMessage('');
 
-    const id = extensionId
-      .replace(/\/$/, '')
-      .split('/')
-      .pop();
+    const isLocalExtension = fs.statSync(extensionId).isDirectory();
```
manojVivek, line 62: @prashantpalikhe Can you please move this logic to the `main` process? We should try and keep `node` usage as less as possible in the `renderer`.

desktop-app/app/components/ExtensionsManager/index.js
```diff
@@ -167,7 +174,9 @@ export default function ExtensionsManager({triggerNavigationReload}) {
 
           <FormHelperText className={styles.extensionsNotice}>
             Note: Only DevTool extensions will work properly, other general
-            browser extensions may not work as intended.
+            browser extensions may not work as intended. Local devtools
+            extensions are also supported. Instead of ID, use absolute path to
```
manojVivek, line 178: @prashantpalikhe If possible can you please add an `open` icon near the `help` icon and let the user select the folder by using the file explorer? 

It will a simple change using the inbuilt `dialog` API, check out a sample usage here - https://github.com/manojVivek/responsively-app/blob/master/desktop-app/app/menu.js#L176

desktop-app/app/components/ExtensionsManager/index.js
prashantpalikhe: Updated. Can you check?

desktop-app/app/components/ExtensionsManager/index.js
prashantpalikhe: Updated. Can you check?

manojVivek:
@prashantpalikhe These looks great, merging it in. Looking forward to more collaboration in the future! 🙌

manojVivek:
@all-contributors Please add @prashantpalikhe to code.

PR #262, opened 2020-07-03
Git ignore .vscode

ziabs:
Git ignore update for .vscode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 .DS_Store
+Thumbs.db
+.vscode


jjavierdguezas:
What do you think about this @manojVivek ? Should we keep that folder in the repo?
I really don't care ...

manojVivek:
@jjavierdguezas  I think keeping it in the repo will be more helpful since most contributors use VSCode as their IDE.

jjavierdguezas:
I'm sorry @manojVivek  I misunderstood your comment and merge this, I reverted it already

PR #265, opened 2020-07-03
increase z-index of devtools container

### LINKED ISSUE #257: DevTools first row disappears when it is open and you go to customize devices
>Version: 0.5.0
Electron: 8.3.0
Chrome: 80.0.3987.165
Node.js: 12.13.0
V8: 8.0.426.27-electron.0
OS: Windows_NT x64 10.0.18362

If you have the dev tools open and you go to customize devices, the dev tools remains there but the first row disappears, so we cannot close it
dev_tools

jjavierdguezas:
resolves #257 
z-index did the trick 😉


manojVivek:
That's great! 😀💪

PR #277, opened 2020-07-05
Alignment of elements in view

vlazaroes:
Previously, elements were not properly aligned with each other.

screenshot


manojVivek:
Thanks for the fix @vlazaroes!

PR #283, opened 2020-07-06
Allow zooming using mousewheel

Grafikart:
## Features

- You can now zoom using CTRL + wheel, 10% per 10% (if your cursor is out of a browser window)
- The zoom level is now displayed in the status bar so we can quickly glance at the zoom level


manojVivek:
This is a great one @Grafikart! 👏 👏

Issue #307, opened 2020-07-12
Add an option in user preference to configure screenshots location


richosr:
The destination for screen shots needs to be configurable.

Writing to the protected desktop is not good practice and is flagged by Anti-Virus software as BAD!

![image](https://user-images.githubusercontent.com/992654/87240428-bc58a380-c443-11ea-8c19-53501fc9ef4a.png)

Steve

manojVivek:
@richosr Thanks for raising this, adding it to the roadmap.

PR #308, opened 2020-07-12
[#307] added option to configure screenshots location.

kirubakarthikeyan:


desktop-app/app/components/UserPreferences/index.js
```diff
@@ -106,6 +107,15 @@ export default function UserPreference({
           />
         </div>
       </div>
+      <div className={cx(commonStyles.sidebarContentSectionContainer)}>
+        <ScreenShotSavePreference
+          screenShotSavePath={
+            userPreferences.screenShotSavePath ||
+            `Desktop/Responsively-Screenshots`
```
manojVivek, line 114: Can you please change this to the full path? So that users can just copy-paste the path if required.

kirubakarthikeyan: Thanks @manojVivek , full path means have to get the whole path from the ipc renderer via event ? below i have attached the screenshot for user preferences section

screenshot

manojVivek: I think you can move the below logic to a util function and use it both here and in `app/menu.js` file too

```
const getDefaultScreenshotpath = () => path.join(
                os.homedir(),
                `Desktop/Responsively-Screenshots`,
                directoryPath
              );
```

desktop-app/app/components/UserPreferences/index.js
```diff
@@ -106,6 +107,15 @@ export default function UserPreference({
           />
         </div>
       </div>
+      <div className={cx(commonStyles.sidebarContentSectionContainer)}>
```
manojVivek, line 110: @kirubakarthikeyan 2 small UI things here:

1. Can you add a label to this input field as `Screenshot Location:`.
2. Move this input section a little down, so that there is some space between the checkbox settings and this one. It would be great if you could add a subtle top border to market the separation.

desktop-app/app/components/UserPreferences/index.js
kirubakarthikeyan: Thanks for the feedback, i've moved the logic into userPreferenceSettings class.

desktop-app/app/components/UserPreferences/index.js
kirubakarthikeyan: I've updated with the neccessary changes, below i've attached the screenshot for the changes i've made.

screenshot

manojVivek:
@kirubakarthikeyan This looks great, merging it now. 🎉 
Looking forward to more collaboration in the future!  🙌

manojVivek:
@all-contributors Please add @kirubakarthikeyan for code.

manojVivek:
@kirubakarthikeyan Two things:

1. Can you let us know your Twitter handle? We would love to give a shout out for your work from our twitter account. 
2. Please join our community on Gitter - https://gitter.im/responsively-app/, for any discussions in the future.

PR #310, opened 2020-07-12
mirroring actions using browser sync

esprush:
#310


desktop-app/app/constants/browserSync.js
```diff
@@ -0,0 +1,5 @@
+export const BROWSER_SYNC_PORT = 12719;
+export const BROWSER_SYNC_VERSION = '2.26.7';
+
+export const BROWSER_SYNC_HOST = `localhost:${BROWSER_SYNC_PORT}`;
+export const BROWSER_SYNC_EMBED_SCRIPT = `https://${BROWSER_SYNC_HOST}/browser-sync/browser-sync-client.js?v=${BROWSER_SYNC_VERSION}`;
```
manojVivek, line 5: @esprush Is the version mandatory here?

desktop-app/app/main.dev.js
```diff
@@ -212,6 +220,7 @@ const createWindow = async () => {
       nodeIntegrationInWorker: true,
       webviewTag: true,
       enableRemoteModule: true,
+      webSecurity: false,
```
manojVivek, line 223: @esprush Any specific reason for adding this?

desktop-app/app/components/WebView/index.js
```diff
@@ -449,51 +456,6 @@ class WebView extends Component {
...
-
-      document.addEventListener(
-        'click',
-        (e) => {
-          if (e.target === window.responsivelyApp.lastClickElement || e.responsivelyAppProcessed) {
-            window.responsivelyApp.lastClickElement = null;
-            e.responsivelyAppProcessed = true;
-            return;
-          }
-          if (window.responsivelyApp.domInspectorEnabled) {
```
manojVivek, line 473: @esprush Removing this will break the inspector, I guess.

desktop-app/app/constants/browserSync.js
esprush: @manojVivek works without the version also. was not sure of the exact implications of removing it, so kept exactly as it was suggested.

desktop-app/app/main.dev.js
esprush: added for experimental purpose. removed this.

desktop-app/app/components/WebView/index.js
esprush: changed @manojVivek . removed only the section where click event was sent.

desktop-app/app/constants/browserSync.js
esprush: @manojVivek works without the version also. was not sure of the exact implications of removing it, so kept exactly as it was suggested.

PR #339, opened 2020-07-20
Browser sync mirroring bug fixes

esprush:


desktop-app/app/service/browserSync.js
```diff
@@ -0,0 +1,14 @@
+import {ipcRenderer} from 'electron';
+import browserSync from 'browser-sync';
+
+let browserSyncOptions;
+
+if (!browserSyncOptions) {
+  browserSyncOptions = ipcRenderer.sendSync('request-browser-sync');
```
manojVivek, line 7: @esprush Can you convert this to `ipcRenderer.invoke` instead of `sendSync`? 
This is going to block the UI thread until the request is processed which is not right.

desktop-app/app/utils/browserSync.js
```diff
@@ -1,17 +1,35 @@
...
+    browserSync.init(
+      {
+        open: false,
+        localOnly: true,
+        https: true,
+        notify: false,
+        ui: false,
+      },
+      (err, bs) => {
+        resolve(bs);
```
manojVivek, line 30: This should be like below:
```
if (err) {
  return reject(err);
}
resolve(bs);
```

Issue #255, opened 2020-07-03
How hide scroll bar of mobile devices when i use mouse?


uzprocode:
<img width="1906" alt="螢幕快照 2020-07-03 下午5 21 38" src="https://user-images.githubusercontent.com/21970517/86454366-b4b54280-bd51-11ea-8010-a4538be7be9b.png">
hide scroll bar.

manojVivek:
@uzprocode This is a good suggestion. Will add it to the roadmap. 🙌

jzabala:
I'll like to take this one!

manojVivek:
Awesome @jzabala. Thanks for the quick PR!

jjavierdguezas:
>Version: 0.9.0
Electron: 9.1.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.18363

@manojVivek  @jzabala  maybe we need to apply css styles to all scrollbars and not only the one of the `body` 
![image](https://user-images.githubusercontent.com/13673443/88825757-3caf3080-d1c8-11ea-99df-f78c23695171.png)

![image](https://user-images.githubusercontent.com/13673443/88826332-e55d9000-d1c8-11ea-8ee0-1dfeb715344e.png)

jzabala:
Yeah, that looks like a better solution. Thanks for pointing it out @jjavierdguezas. I can create the PR for that.

PR #364, opened 2020-07-29
fix(desktop-app): on mobile devices hide scrollbar on all elements

jzabala:
As suggested on issue #255 comment https://github.com/manojVivek/responsively-app/issues/255#issuecomment-665766060 a better approach to hiding the scrollbar on mobile devices is to hide it on all elements.


manojVivek:
Thanks for the quick fix @jzabala

PR #373, opened 2020-08-03
file watching using browser sync

esprush:
@manojVivek @jjavierdguezas hot reloading for local files using browserSync. added support for js,html and css.


desktop-app/app/utils/browserSync.js
```diff
@@ -1,15 +1,38 @@
...
 export async function initBrowserSync() {
-  if (!browserSync.active) {
+  if (!browserSyncEmbed.active) {
     await initInstance();
   }
 }
 
+export function watchFiles(fileDir) {
+  filesWatcher = browserSyncEmbed
+    .watch([`${fileDir}/*.html`, `${fileDir}/**/**.js`])
```
jjavierdguezas, line 15: @esprush 
If we are watching all `*.js` and all `*.css` in all sub folders, why not to do the same with `*.html`?

desktop-app/app/main.dev.js
```diff
@@ -305,30 +307,19 @@ const createWindow = async () => {
     onResize();
   });
 
-  const watcher = new chokidar.FSWatcher();
```
jjavierdguezas, line 308: should we uninstall `chokidar` package right?

desktop-app/app/utils/browserSync.js
manojVivek: I think we should only watch the currently loaded `.html` file, as the other HTML files are not related to the current page.

manojVivek: @esprush  @jjavierdguezas Merging this now. Let's address any concerns in a new PR.

PR #398, opened 2020-08-16
URI-encode-on-open-html-file

aUsefulMethod:
I'm able to open .html files with # in name when opening with responsively. I've tested a few other reserved characters and it seems to be working as well.

### DIFF
--- a/desktop-app/app/main.dev.js
+++ b/desktop-app/app/main.dev.js
@@ -91,7 +91,10 @@ const openWithHandler = filePath => {
     filePath != null &&
     (filePath.endsWith('.html') || filePath.endsWith('.htm'))
   ) {
-    const url = `file://${filePath}`;
+    // code to encode
+    // file URI
+    p1 = encodeURIComponent(filePath);
+    const url = `file://${filePath + p1}`;
     fileToOpen = url;
     return true;
   }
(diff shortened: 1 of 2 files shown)


desktop-app/app/main.dev.js
```diff
@@ -91,7 +91,10 @@ const openWithHandler = filePath => {
     filePath != null &&
     (filePath.endsWith('.html') || filePath.endsWith('.htm'))
   ) {
-    const url = `file://${filePath}`;
+    // code to encode
+    // file URI
+    p1 = encodeURIComponent(filePath);
+    const url = `file://${filePath + p1}`;
```
manojVivek, line 97: @aUsefulMethod Isn't `filePath + p1` going to repeat the path twice?

manojVivek:
@aUsefulMethod Can you please attach a screenshot of how the address bar looks after the encoding?

jzabala:
Thanks for the PR @aUsefulMethod 🙂

I have a question about what problem is this PR solving? 

From what I see Chromium is already encoding the path.

Aug-20-2020 11-26-47

desktop-app/app/main.dev.js
mrfelfel: maybe not work ... encodeURIComponent  encode all of string :D

aUsefulMethod:
I’m currently in school, I’ll close this PR when I get a chance.

jjavierdguezas:
Hi @aUsefulMethod we are going with #415, we think that is a better approach. Thanks anyway and we are looking forward to more collaboration in the future! 🙌

PR #422, opened 2020-08-26
Keep inactive devices order on device management

### LINKED ISSUE #234: Inactive Devices list keep reordering itself
>Version: 0.5.0
Electron: 8.3.0
Chrome: 80.0.3987.165
Node.js: 12.13.0
V8: 8.0.426.27-electron.0
OS: Windows_NT x64 10.0.18362

If this is intended, then just close this issue... but I personally think it is a bit weird

When managing the devices, no matters how I drag and drop an item in the Inactive Devices list, it keeps reordering itself. Maybe I want to keep an order on that list or I made a mistake and moved the wrong device to inactives, after a drag&drop action the moved device should be where I dropped it, but that is not always the case
This only happens in the Inactive Devices list

device list

jjavierdguezas:
resolves #234 
also this PR fix some style issues when adding new devices

### DIFF
--- a/desktop-app/app/components/DeviceManager/AddDevice/index.js
+++ b/desktop-app/app/components/DeviceManager/AddDevice/index.js
@@ -43,9 +43,13 @@ import {lightIconsColor, themeColor} from '../../../constants/colors';
 
 const useStyles = makeStyles(theme => ({
   fab: {
-    position: 'absolute',
+    position: 'absolute !important',
     top: theme.spacing(10),
     right: theme.spacing(3),
+    color: '#fff !important',
+    backgroundColor: `${themeColor} !important`,
+    borderRadius: '24px !important',
+    padding: '0 16px !important',
   },
   extendedIcon: {
     marginRight: theme.spacing(1),
--- a/desktop-app/app/components/DeviceManager/index.js
+++ b/desktop-app/app/components/DeviceManager/index.js
@@ -45,9 +45,26 @@ export default function DeviceManager(props) {
       acc[val.id] = val;
       return acc;
     }, {});
-    const inactiveDevices = props.browser.allDevices.filter(
-      device => !activeDevicesById[device.id]
-    );
+
+    const currentInactiveDevicesById = devices.inactive.reduce((acc, val) => {
+      acc[val.id] = val;
+      return acc;
+    }, {});
+
+    const devicesById = props.browser.allDevices.reduce((acc, val) => {
+      acc[val.id] = val;
+      return acc;
+    }, {});
+
+    const inactiveDevices = [
+      ...props.browser.allDevices.filter(
+        device =>
+          !activeDevicesById[device.id] &&
+          !currentInactiveDevicesById[device.id]
+      ),
+      ...devices.inactive.filter(device => devicesById[device.id]),
+    ];
+
     setDevices({active: activeDevices, inactive: inactiveDevices});
   }, [props.browser.devices, props.browser.allDevices]);
 
@@ -75,9 +92,21 @@ export default function DeviceManager(props) {
       source.droppableId === 'inactive'
         ? devices.inactiveFiltered[source.index]
         : sourceList[source.index];
+
+    let idx = destination.index;
+
+    if (destination.droppableId === 'inactive') {
+      idx =
+        destination.index < devices.inactiveFiltered.length
+          ? devices.inactive.findIndex(
+              d => d.id === devices.inactiveFiltered[destination.index].id
+            )
+          : devices.inactive.length;
+    }
+
     sourceList.splice(sourceList.indexOf(itemDragged), 1);
 
-    destinationList.splice(destination.index, 0, itemDragged);
+    destinationList.splice(idx, 0, itemDragged);
 
     updateDevices(devices);
   };
--- a/desktop-app/app/components/DeviceManager/DeviceList/index.js
+++ b/desktop-app/app/components/DeviceManager/DeviceList/index.js
@@ -59,6 +59,7 @@ export default function DeviceList({
                   endAdornment: (
                     <InputAdornment>
                       <IconButton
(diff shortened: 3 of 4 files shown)


desktop-app/app/components/DeviceManager/index.js
```diff
@@ -45,9 +45,26 @@ export default function DeviceManager(props) {
...
+      return acc;
+    }, {});
+
+    const devicesById = props.browser.allDevices.reduce((acc, val) => {
+      acc[val.id] = val;
+      return acc;
+    }, {});
+
+    const inactiveDevices = [
+      ...props.browser.allDevices.filter(
```
jzabala, line 60: During testing I saw this expression only returned something during the first render. If that is the case all the work done to generate this value in subsequent renders is unnecessary. I would recommend to move that work to a `useMemo` that only runs once.

jjavierdguezas: @jzabala Are you considereing the scenario where the user adds or removes devices ?, In those scenarios the `browser.allDevices` changes right?

desktop-app/app/components/DeviceManager/AddDevice/index.js
```diff
@@ -43,9 +43,13 @@ import {lightIconsColor, themeColor} from '../../../constants/colors';
 
 const useStyles = makeStyles(theme => ({
   fab: {
-    position: 'absolute',
+    position: 'absolute !important',
```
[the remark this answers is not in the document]
jjavierdguezas: I dont know why but when you try to Add a new custom device the button style get override and it moves to the bottom of the modal. With those `!important`s the style of the button is always the same and it does not move.

desktop-app/app/components/DeviceManager/index.js
jzabala: Yeah, but it is added to the active devices making the expression false at line 62.

desktop-app/app/components/DeviceManager/AddDevice/index.js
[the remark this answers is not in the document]
jzabala: Oh yeah, you are right. It disappears when the modal is open. That is interesting 🤔

desktop-app/app/components/DeviceManager/index.js
jjavierdguezas: You can add new devices to 'inactive' list directly too

jjavierdguezas: image

desktop-app/app/components/DeviceManager/AddDevice/index.js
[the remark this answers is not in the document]
jjavierdguezas: btw, that is the style issue I fix in this PR: keep the Add button in place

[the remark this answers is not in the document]
jzabala: Got it!

desktop-app/app/components/DeviceManager/index.js
jzabala: Is it an expected behavior that Custom devices in the inactive list always get to the top when I close and open the app?

jzabala:
@jjavierdguezas 

Is it an expected behavior that Custom devices in the inactive list always get to the top when I close and open the app?

jjavierdguezas:
> @jjavierdguezas
> 
> Is it an expected behavior that Custom devices in the inactive list always get to the top when I close and open the app?

I think it is,
`browser.allDevices` is filled calling [`getAllDevices`](https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/constants/devices.js#L76-L102) 
https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/reducers/browser.js#L335

which put them in that order @manojVivek 
https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/constants/devices.js#L101

jzabala:
> > @jjavierdguezas
> > Is it an expected behavior that Custom devices in the inactive list always get to the top when I close and open the app?
> 
> I think it is,
> `browser.allDevices` is filled calling [`getAllDevices`](https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/constants/devices.js#L76-L102)
> https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/reducers/browser.js#L335
> 
> which put them in that order @manojVivek
> https://github.com/responsively-org/responsively-app/blob/1366d77eda5c3ad31fde93098979bed0bafba59b/desktop-app/app/constants/devices.js#L101

Thanks for the quick response @jjavierdguezas 

Now that I check it better I see that the only list that keeps the order after reset is the list of active devices. I ordered both lists before restating the app and as you can see in the image bellow only the active list keep the changes.

Screen Shot 2020-08-26 at 3 45 53 PM

Screen Shot 2020-08-26 at 3 47 06 PM

jjavierdguezas:
> ... I see that the only list that keeps the order after reset is the list of active devices. I ordered both lists before restating the app and as you can see in the image bellow only the active list keep the changes....

I think that we need to decide this @manojVivek @esprush @jzabala what do you think about it? Should we discuss it on Slack?

manojVivek:
> Is it an expected behavior that Custom devices in the inactive list always get to the top when I close and open the app?

Yes, I make it like that, so that the custom devices are on the top and easier for the user to use them.

manojVivek:
@jzabala @jjavierdguezas I'm merging this in, please add new issue if there are any outstanding items.

PR #425, opened 2020-08-26
Feature/Screenshot v2

### LINKED ISSUE #424: Screenshots v2
The main reason behind this new screenshot mechanism is to get away with the logic of capturing portions of the screens and then stitching it all together to form the complete image.

V2:
1. Instead of capturing different sections of the page by scrolling to that location, we will temporarily increase the height and width of the `webview` to match the documents dimensions. This way `webContents.capturePage()` will capture the whole page in a single shot.
2. When increasing the height of the `webview` to the document's dimensions, it is possible that some portion of the page will go outside the user's viewport and the overflowing portion will be captured blank by chromium engine. So in order to mitigate this, we will temporarily reduce the `zoomfactor` to `0.01` so that the whole `webview` will fit into the viewport.
3. Previously we were capturing and saving the screenshots in `png` format, after doing some experiments I realized that `jpg` format is much faster than `png`s so changed the image format to `jpg`.

This change might not work as intended for some sites(like google.com) that adjust the size of their contents based on the viewport, so made added an option in the user preference to choose the desired capture mechanism.

manojVivek:
Please refer to #424's description for what is being changed. 

No changes in the UI, the screenshot time has reduced by at least half in all the tests that I did.

Fix #424


desktop-app/app/components/UserPreferences/index.js
```diff
@@ -128,6 +129,33 @@ export default function UserPreference({
...
+                      : SCREENSHOT_MECHANISM.V1
+                  )
+                }
+                name="Use improved screenshot mechanism"
+                color="primary"
+              />
+            }
+            label={
+              <span className={cx(styles.preferenceName)}>
+                Use improved screenshot mechanism
```
jjavierdguezas, line 154: Maybe change this to `Use improved mechanism`? "screenshot" is implicit and it took just one line un the drawer
image
vs
image

desktop-app/app/components/WebView/screenshotUtil.js
```diff
@@ -8,13 +8,17 @@ import {promisify} from 'util';
...
 import path from 'path';
 import fs from 'fs-extra';
+import Jimp from 'jimp';
 import PromiseWorker from 'promise-worker';
 import NotificationMessage from '../NotificationMessage';
 import {userPreferenceSettings} from '../../settings/userPreferenceSettings';
 import {type Device} from '../../constants/devices';
 import {captureOnSentry} from '../../utils/logUtils';
+import {SCREENSHOT_MECHANISM} from '../../constants/values';
+import mutexify from 'mutexify/promise';
```
jjavierdguezas, line 18: you need to add `mutexify` to `package,json`

jjavierdguezas:
WOW this is FAST!!!

desktop-app/app/components/UserPreferences/index.js
manojVivek: Yeah, this is much better, I'll change it. 👍

desktop-app/app/components/WebView/screenshotUtil.js
manojVivek: Yeah, not sure how it got missed. Will add it.

manojVivek:
@jjavierdguezas I have addresssed the comments.

jjavierdguezas (approved):

PR #426, opened 2020-08-27
add permission request prompt.

### LINKED ISSUE #370: Add a way to show the user-permission requests by the webpage.
Currently, the user-permission requests for showing notifications/access to geolocation are being ignored and not shown to the app user. 
Add a mechanism to show a prompt to the user about the request.

jjavierdguezas:
resolve #370
This is how I imagine it, highly inspired by Google Chrome behavior, suggestions are welcome
permissions

I think this is the minimum for this feature, we have to think about persisting user decisions during app restarts, also find a way to manage per website permissions to revoke or change a previous decision (as chrome do).


manojVivek:
> This is how I imagine it, highly inspired by Google Chrome behavior, suggestions are welcome

@jjavierdguezas This looks great. 👍 

> I think this is the minimum for this feature, we have to think about persisting user decisions during app restarts, also find a way to manage per website permissions to revoke or change a previous decision (as chrome do).

Yeah, this is good for now. We can improve it later when users feel the need for it.

desktop-app/app/components/NavigationControls/index.js
```diff
@@ -58,7 +59,10 @@ class NavigationControls extends Component {
           <Tooltip title="Reload">
             <div
               className={commonStyles.flexAlignVerticalMiddle}
-              onClick={this.props.triggerNavigationReload}
+              onClick={() => {
+                notifyPermissionToHandleReloadOrNewAddress();
```
manojVivek, line 63: @jjavierdguezas Should we move `notifyPermissionToHandleReloadOrNewAddress()` to the ADDRESS_CHANGE action handler instead of the individual components? 

That way we won't miss this when some other component does the address change in the future.

desktop-app/app/components/UserPreferences/index.js
```diff
@@ -5,12 +5,46 @@ import Checkbox from '@material-ui/core/Checkbox';
...
+    };
+  },
+  input: selectStyles => ({...selectStyles}),
+  placeholder: selectStyles => ({...selectStyles}),
+  singleValue: (selectStyles, {data}) => ({...selectStyles, color: 'white'}),
+  menu: selectStyles => ({...selectStyles, background: '#4b4b4b', zIndex: 100}),
+};
+
+const permissionsOptions = [
+  {value: 'Allow always', label: 'Allow always'},
```
manojVivek, line 44: Can you move the string values to a constants file and refer from there?

For a similar use-case in #425 PR, I have added these to  `src/contansts/values.js` file.

desktop-app/app/main.dev.js
```diff
@@ -319,6 +321,103 @@ const createWindow = async () => {
     onResize();
   });
 
+  const getPermissionSettingPreference = () => {
```
manojVivek, line 324: @jjavierdguezas Can you please move this function to `permissionUtils.js` and import from there?

manojVivek (changes requested):

desktop-app/app/components/NavigationControls/index.js
jjavierdguezas: yes, this is an ugly solution.  I used the `NEW_ADDRESS` action first, but for this feature you need to know when the address changed **before** the webview reloads and starts asking for permission.  After lots of testing, changes, and retries I settled on this because sometimes (go back and forward for example) the `NEW_ADDRESS` action has a long delay and as a result of that the component was handling new permission requests before find out about reloads, breaking the logic.  The (ugly) solution was to send a new and specific event as soon as possible when I know the address changed or the page reloads 😓. I really don't know how to avoid this, and yes any future change that triggers a page reload or address change needs to call `notifyPermissionToHandleReloadOrNewAddress()` function

desktop-app/app/main.dev.js
jjavierdguezas: done

desktop-app/app/components/UserPreferences/index.js
jjavierdguezas: done

desktop-app/app/reducers/browser.js
```diff
@@ -195,9 +197,17 @@ function _getActiveDevices() {
...
-  );
+  const fromSettings = settings.get(USER_PREFERENCES);
+  if (fromSettings == null)
+    return {
+      removeFixedPositionedElements: true,
+      permissionManagement: PERMISSION_MANAGEMENT_OPTIONS.ALLOW_ALWAYS,
+    };
+
+  if (fromSettings.permissionManagement == null)
+    fromSettings.permissionManagement = PERMISSION_MANAGEMENT_OPTIONS.ALLOW_ALWAYS;
```
manojVivek, line 208: @jjavierdguezas Can you set this as the default value in the https://github.com/responsively-org/responsively-app/blob/master/desktop-app/app/settings/migration.js#L25 file? 

So that we can avoid this fallback cases.

jjavierdguezas: done

PR #441, opened 2020-08-31
Assume HTTP for localhost subdomains

### LINKED ISSUE #433: Should it assume HTTP for localhost?
## Current Behavior
1. Type `taeluf.localhost` (or your own localhost subdomain)
2. [enter]
3. address bar becomes: `https://taeluf.localhost`

## Desired Behavior
When `.localhost` tld is used, route to `http://taeluf.localhost`

## My thoughts
All my sites use `HTTPS`, but not in local development. Firefox assumes `http://`, and I think other browsers do to. Or maybe there is some "dns" type lookup that determines it?  

I'm running an Apache server. In playing with `npm run start`, that also makes an `http` localserver (I did zero configuration & have very little experience with npm).

I definitely think `https://` is better for the web & should be assumed there.

I'm interested in **trying** to contribute a "fix" for this if y'all like it.

jjavierdguezas:
resolves #433

### DIFF
--- a/desktop-app/app/components/Addressinput/index.js
+++ b/desktop-app/app/components/Addressinput/index.js
@@ -291,10 +291,21 @@ class AddressBar extends React.Component<Props> {
     });
   };
 
+  _hostnameCharHints = [':', '/', '#', '&', '?'];
+  _inferHostname = (address: string) =>
+    this._hostnameCharHints.reduce(
+      (curr, char) => curr.split(char)[0],
+      address
+    );
+
   _normalize = (address: string) => {
     if (address.indexOf('://') === -1) {
       let protocol = 'https://';
-      if (address.startsWith('localhost') || address.startsWith('127.0.0.1') || address.split(':')[0].split('?')[0].endsWith('.localhost')) {
+      if (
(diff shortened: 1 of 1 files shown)


desktop-app/app/components/Addressinput/index.js
```diff
@@ -294,7 +294,7 @@ class AddressBar extends React.Component<Props> {
   _normalize = (address: string) => {
     if (address.indexOf('://') === -1) {
       let protocol = 'https://';
-      if (address.startsWith('localhost') || address.startsWith('127.0.0.1')) {
+      if (address.startsWith('localhost') || address.startsWith('127.0.0.1') || address.split(':')[0].split('?')[0].endsWith('.localhost')) {
```
manojVivek, line 297: @jjavierdguezas  Will this match if the entered URL is something like: `http://test.localhost/`?

jjavierdguezas: mmmm missed that! 😅

jjavierdguezas: done

Issue #423, opened 2020-08-26
Improve Manage Devices view UX


jzabala:
1. `CommandOrControl+F` (depending on the platforms) in this view should open and focus the search input.

![Aug-26-2020 08-51-42](https://user-images.githubusercontent.com/1315054/91305931-d04d3000-e779-11ea-9052-082d7bdf0f29.gif)

2. `Escape` with the search input focused should close it and not the whole view.

![Aug-26-2020 08-53-07](https://user-images.githubusercontent.com/1315054/91306044-0094ce80-e77a-11ea-8917-9a7ea184e0bb.gif)

maxwillkelly:
I've managed to implement the Command/Control to focus input, should be able to implement the second tweak and make a pull request soon.

jjavierdguezas:
great @maxwillkelly  assign this to you then

PR #400, opened 2020-08-17
Feature/find text

esprush:
Screenshot from 2020-08-17 23-27-32

PR #443, opened 2020-09-02
Improve Manage Devices view UX

maxwillkelly:
This commit implements the two changes suggested under Issue #423. This means in the Device Manager, Control/Command + F will now open and focus the search input for Inactive devices. Also, the Escape key no longer closes the device manager and only closes the search field when open. 

I originally envisioned a solution where the Escape key would close the device manager if the search input was not focussed but that proved more complicated than I could manage. 

If closing the Device Manager by pressing Escape is deemed more necessary than closing the search field, this can easily by reenabled though a change to one-line.


jjavierdguezas:
I am concerned that this conflicts with #400 (use of `mod+F`)
@esprush @manojVivek @jzabala

manojVivek:
> I am concerned that this conflicts with #400 (use of `mod+F`)

Yeah, #400's shortcut takes the priority.

This shortcut should only be active when the device manager is open. Maybe something at the component level can help instead of the global shortcut?

jjavierdguezas:
> This shortcut should only be active when the device manager is open. Maybe something at the component level can help instead of the global shortcut?

I think that maybe to move the shortcut in #400 from menu to our custom shortcut registration in `Root` component, then put an `if` like
```js
registerShortcut(
      {
        id: 'Find',
        title: 'Find',
        accelerators: ['mod+f'],
      },
      () => {
        if ( MANAGING_DEVICES )
             store.dispatch(onSearchDevicesActivated()); // search in inactive device list
        else
            store.dispatch(onFindInDevicesActivated());  // search in webview contents
      },
      true
    );
```

maxwillkelly:
No problem, would the best approach be to keep the pull requests separate and try to avoid conflicts for later or to merge  @esprush's branch into the mine now?

jzabala:
> ```js
>  search in inactive device list
> ```

I don't think we need electron shortcuts for this one. We can just add a keypress event on the div and check for the [metaKey][metakey] (on Mac) or [ctrlKey][ctrlkey](Window, Linux). If we are 

[metakey]: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey
[ctrlkey]: https://www.w3schools.com/JSREF/event_ctrlkey.asp

jzabala:
> No problem, would the best approach be to keep the pull requests separate and try to avoid conflicts for later or to merge @esprush's branch into the mine now?

I think it would be better to keep them separate and since this is not a global shortcut not to use electron for it. 

https://github.com/responsively-org/responsively-app/pull/443#issuecomment-686856933

maxwillkelly:
Ok, should have it reworked by Monday

maxwillkelly:
Those commits make the shortcut for Command/Control + F component-based. Escape also now closes the device manager when the search box is not focused.

desktop-app/app/components/DeviceManager/DeviceList/index.js
```diff
@@ -10,91 +10,109 @@ import DeviceItem from '../DeviceItem';
...
-  devices,
-  enableFiltering,
-  onFiltering,
-  enableCustomDeviceDeletion,
-  deleteDevice,
-}) {
-  const [searchOpen, setSearchOpen] = useState(false);
-  const [searchText, setSearchText] = useState('');
-  const [filteredDevices, setFilteredList] = useState(devices);
+const DeviceListHandler = props => {
```
jzabala, line 13: Why all these changes related to the filtered devices?

maxwillkelly: Fair question. It's because the DeviceManager (where we want to get a keypress) and the DeviceList (containing a search box) are separate components. This means that in order to focus the search box within the DeviceManager we have to have a reference to the DeviceList. The problem is you can't use a ref attribute on a functional component meaning I had to convert the DeviceList into a class component. Because I had trouble getting device filtering working using the class component, I split the device filtering into its own functional handler, meaning most of the code within it, is basically what was there before.

jjavierdguezas: @jzabala @manojVivek  should we go with this?
@maxwillkelly please solve the conflicts by merging `master` into this branch

maxwillkelly: No problem will get that done soon

jjavierdguezas:
Hi @maxwillkelly, are you still working on this? just checking...

maxwillkelly:
@jjavierdguezas Sorry life got in the way. The code has probably changed too much for there to be much salvageable here, so I will close the pull request.

PR #454, opened 2020-09-16
Address bar light mode color fix

manojVivek:


jjavierdguezas (approved):
looks good @manojVivek

PR #465, opened 2020-09-20
Theme preference decoupled from toolbar option

manojVivek:


jjavierdguezas:
Looks good @manojVivek , merging it

Issue #388, opened 2020-08-12
Persistent hover effect when Toggle Touch Mode button is activated.


jzabala:
As described on https://github.com/responsively-org/responsively-app/issues/232#issuecomment-652340907 by @jjavierdguezas 

> When you click the Toggle Touch Mode button and Touch Mode is on, moving the mouse to the next button (left or right) will make it look hovered until you toggle Touch Mode again

![touchmodeissue2](https://user-images.githubusercontent.com/13673443/86234807-c521bd80-bb97-11ea-87ab-449752c6d93d.gif)

_Note: moving this to its own issue since it can be worked independently of the other problems reported on #232_

JLambertazzo:
Found a fix for the persistent hover by changing the toggle button's event handling!

jjavierdguezas:
Great @JLambertazzo! I'm assing it to you. Thanks!

PR #467, opened 2020-09-21
Fixed persistent hover effect

JLambertazzo:
This is a fix to issue https://github.com/responsively-org/responsively-app/issues/388

The cause of the bug was that the state of the pointer (mouse/touch) was updated as the pointer left the button. With this fix, the state is updated on mouse movement within the button, giving it a more responsive feel.

### DIFF
--- a/desktop-app/app/components/ToggleTouch/index.js
+++ b/desktop-app/app/components/ToggleTouch/index.js
@@ -51,12 +51,15 @@ export default function ToggleTouch({iconProps}) {
       className={cx(commonClasses.icon, {
         [commonClasses.iconSelected]: isTouchMode,
       })}
+      onPointerEnter={() => {
+        setHover(true);
+        setTooltip(true);
+      }}
       onPointerLeave={() => {
         setHover(false);
         setTooltip(false);
       }}
       onPointerMove={() => {
-        setTooltip(true);
         if (!hasPendingState) {
           return;
         }
--- a/desktop-app/app/components/ToggleTouch/index.js
+++ b/desktop-app/app/components/ToggleTouch/index.js
@@ -26,7 +26,7 @@ export default function ToggleTouch({iconProps}) {
 
   const handleToggleTouch = () => {
     setIsTouchMode(!isTouchMode);
-    setTooltip(isTouchMode);
+    setTooltip(!isTouchMode);
     if (isHover) {
       return setPendingState(true);
     }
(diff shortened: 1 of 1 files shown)


jjavierdguezas:
Hi @JLambertazzo this looks good! thanks
It is posible to hide the tooltip too? it keeps open even when the pointer is far away
image

jjavierdguezas:
@JLambertazzo can you please make a comment in #388 in order to assign it to you (We can't if you haven't interacted with the issue.)

JLambertazzo:
I can take a look at the tooltip issue!

JLambertazzo:
The problem with the tooltip sticking around was happening because, when touch mode is entered the mouse stays hovered on the toggle button until user is no longer in touch mode. I fixed the issue by hiding the tooltip while in touch mode!
fix

jjavierdguezas:
@JLambertazzo  others buttons always shows tooltip on hover, in order to be consistent, why don't you try a combination of the `open` prop of `<Tooltip ...>` and `onMouseEnter` , `onMouseLeave`, `onPointerEnter` and `onPointerLeave` events? Just an idea
`onXXXXEnter => open = true`
`onXXXXLeave => open = false`
what do you think?

JLambertazzo:
Unfortunately once in touch mode hover is no longer triggered by hovering the cursor over, rather you would need to click and hold to have the same hover effect. These are the results I got when trying to listen and log any focus/pointer/mouse/touch events, which showed me that in touch mode moving the mouse doesn't trigger any of these events.

prob

I was however able to implement consistent behavior for the touch mode icon with all other icons. The other icons would show tooltips in touch mode when you would click and hold them, and now the toggle touch mode button does the same. As for the persistent hover issue, the tooltip sticking around, those issues are fixed and even some of the inconsistency is dealt with. However, the inconsistent behavior between toggle buttons actually seems to be a much deeper issue, and should therefore have it's own issue thread. I'll join the slack to discuss the problem further, unless there is a better place to do so.

jjavierdguezas:
Sorry @JLambertazzo  now there is a weird behavior with the tooltip
touch
If the pointer/mouse events don't fire on touch mode I prefer your initial solution where there is no tooltip on touch mode
`touchMode = true => tooltip = ""`
I rolled back to your initial solution,
Now I merging this, thanks

jjavierdguezas:
@JLambertazzo Thanks so much for your code contribution.
Looking forward to more collaboration in the future! 💪🏻

jjavierdguezas:
@all-contributors Please add @JLambertazzo for code.

jjavierdguezas:
@JLambertazzo Do you might sharing your Twitter handle? We would love to send a shout out for your work on twitter. :)
And please join our slack workspace for any future discussions: https://join.slack.com/t/responsively/shared_invite/zt-haoieftz-IsMw64H6jXC23pJ16ROLzw :)

JLambertazzo:
That would be great, thank you! My twitter handle is @JLambertazzo. I'll reach out on slack.

jjavierdguezas:
> That would be great, thank you! My twitter handle is @JLambertazzo. I'll reach out on slack.

cc: @manojVivek

PR #471, opened 2020-09-22
Refactor remove routing dependencies

sidthesloth92:
What does this PR contain:

- Removed `react-router`, `react-router-dom`, `connected-react-router`, `history` packages from package.json
- Removed code usages of the above-mentioned packages
- Renamed `Routes.js` to `AppContent.js`

**Tests**
Ran the app and it worked fine.. :)


desktop-app/app/Routes.js
```diff
@@ -1,5 +1,4 @@
 import React, {Fragment} from 'react';
```
sidthesloth92, line 1: @manojVivek I think we should rename this as `App.js`, move the app component from `Root.js` here. That would be much cleaner in two ways, `Root.js` will have only one component and `Routes.js` doesn't make any sense since we have removed routing.

manojVivek: Yes this makes sense, please do it.

sidthesloth92: Tried to do it, but `App` component uses a hook and `Root` is a class component. This will lead to further refactor beyond the scope of this PR. So renamed `Routes.js` to `AppContent.js` for now.

manojVivek: ok, that's cool!

manojVivek (approved):
This looks great! thank you so much for your efforts. 

Looking forward to more like this in the future! 🎉

manojVivek:
@all-contributors Please add @sidthesloth92 for code.

manojVivek:
@sidthesloth92 Do you might sharing your Twitter handle? We would love to send a shout out for your work on twitter. :)

PR #485, opened 2020-10-01
Feature: Show/Hide toolbars

sidthesloth92:
**What's in the pull request**
Added buttons to the top and side toolbar to hide/show them.

**Tested Scenarios**
- Toggling of both the toolbar buttons in various scenarios.
- Tried dark and light mode.

**Help needed:**
**I don't have a windows device. So I'd appreciate if someone can check out the branch and test it in Window and Linux.**

**Screenshots/Video**
zenmode

**Dark Mode**
image

**Light Mode**
image


sidthesloth92:
@manojVivek I found one bug. What is the expected behaviour here?
image

I believe we should not show the blip at all when the extended menu is open?

manojVivek:
> I believe we should not show the blip at all when the extended menu is open?

Yes Dinesh, I think so too. We can just hide it when the sidebar is open.

manojVivek:
> **Screenshots/Video**
> zenmode

@sidthesloth92 I think we should flip the icon when it is closed so that the arrow indicates the direction of movement.

desktop-app/app/components/Header/index.js
```diff
@@ -12,12 +12,17 @@ import NavigationControlsContainer from '../../containers/NavigationControlsCont
...
+import ZenButton from '../ZenButton';
 
-const Header = () => {
+const Header = props => {
   const classes = useStyles();
 
   return (
-    <div className={classes.container}>
+    <div
+      className={`${classes.container} ${
```
manojVivek, line 22: @sidthesloth92 Can you please make use of the `classnames` module here? 
That will make it more readable.

desktop-app/app/components/Header/index.js
```diff
@@ -53,22 +58,45 @@ const Header = () => {
...
 
 const useStyles = makeStyles(theme => ({
   container: {
+    position: 'relative',
     background: theme.palette.background.l1,
     display: 'flex',
     flexDirection: 'column',
     width: '100%',
-    padding: os.platform() === 'darwin' ? '20px 0 5px' : '0 0 0',
+    padding: os.platform() === 'darwin' ? '0 0 5px' : '0 0 0',
```
manojVivek, line 75: Removing the padding is going to make the OS's minimize/maximize buttons overlap on the toolbar right?

manojVivek: Nevermind, just noticed the `Toolbar` component.

desktop-app/app/components/Toolbar/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import {makeStyles} from '@material-ui/core/styles';
+
+/**
+ * Application toolbar that appears at the top of a window.
+ */
+const Toolbar = () => {
```
manojVivek, line 7: @sidthesloth92 Can we name it something like a Horizontal Spacer? As it is not technically a toolbar.

manojVivek (changes requested):
Just a couple of comments, looks great otherwise! 👏

desktop-app/app/components/Header/index.js
sidthesloth92: Sure will do. Wasn't aware of the module.. :D

desktop-app/app/components/Toolbar/index.js
sidthesloth92: Totally agreed..😭

jjavierdguezas (commented):
@sidthesloth92 How does permissions popup looks like on 'zen mode'?

jjavierdguezas:
> @sidthesloth92 How does permissions popup looks like on 'zen mode'?

I saw it, it looks good

jjavierdguezas:
@sidthesloth92 @manojVivek 
Should we hide the handles on 'zen mode' too? (and show it when hovering borders)
Should we expand the area of the devices on 'zen mode'? (and use the space left by hidden toolbars)

desktop-app/app/components/Header/index.js
sidthesloth92: Done.

desktop-app/app/components/Toolbar/index.js
sidthesloth92: Done.

sidthesloth92:
Updated GIF
zen-mode-2

manojVivek:
> Should we hide the handles on 'zen mode' too? (and show it when hovering borders)

@jjavierdguezas I think hiding it completely might puzzle the users who are not aware of it. So it is better to show the handles when hidden.

PR #486, opened 2020-10-03
Electron 9 minor version upgrade

manojVivek:


sidthesloth92 (approved):

sidthesloth92:
@manojVivek Do we have test suites to ensure functionality is not broken?

manojVivek:
> @manojVivek Do we have test suites to ensure functionality is not broken?

@sidthesloth92 Unfortunately we don't have unit test suites. 😓 
I have manually tested most of the workflow. Also checked all the release notes from the previous version to the current one and there is no breaking change in any of those. So we should be good, I believe.

sidthesloth92:
That's some testing..👏 The problem though is that everyone doesn't have same app knowledge like you.. :( (Including me) I was little scared when I committed code to be honest.. 😅 Should we create an issue? [Specton](https://www.electronjs.org/spectron) seems to be promising.

manojVivek:
> Should we create an issue? [Specton](https://www.electronjs.org/spectron) seems to be promising.

Yep, we should definitely start building a test suite. I'll create an issue from your comment.

jjavierdguezas (approved):

PR #501, opened 2020-10-05
Add design mode

### LINKED ISSUE #491: [Feature] Turn on Design Mode
Add buttons (per devices and for all) that toggle Design Mode in the device
This should be in the ⋮ button (_kebab menu?_) since it is not a usual action

This allows users to edit HTML directly without dev tools. Sadly I don't think it can be synchronized across devices, because events are disabled on Design Mode (but it is a good topic to investigate)

designMode

**Hints:**
`document.designMode = "on"`
https://developer.mozilla.org/es/docs/Web/API/Document/designMode

jjavierdguezas:
Resolves #491

designMode

(we need to think about to move some actions to a kebab menu ` ⋮ ` on the main toolbar, it is getting too crowded there )


manojVivek:
> we need to think about to move some actions to a kebab menu `⋮` on the main toolbar, it is getting too crowded there

@jjavierdguezas Maybe we should move the `Full Screenshot` option to the kebab menu? What do you think?

desktop-app/app/actions/browser.js
```diff
@@ -698,6 +716,23 @@ export function onDeviceMutedChange(deviceId, isMuted) {
...
+    const {
+      browser: {allDevicesInDesignMode},
+    } = getState();
+    const next = !allDevicesInDesignMode;
+    pubsub.publish(TOGGLE_DEVICE_DESIGN_MODE_STATE, [{designMode: next}]);
+    dispatch(toggleAllDevicesDesignMode(next));
+  };
+}
+
+export function onDeviceDesignModeChange(deviceId, designModeOn) {
```
sidthesloth92, line 730: Since the functionality is a toggle of a given device's design mode, it should just take the `deviceId` as the parameter and use the state in the reducer to toggle the flag. This ensures that the consumer can only toggle as the functionality implies.

desktop-app/app/actions/browser.js
```diff
@@ -185,6 +188,21 @@ export function toggleDeviceMuted(deviceId, isMuted) {
...
 }
 
+export function toggleAllDevicesDesignMode(allDevicesInDesignMode) {
+  return {
+    type: TOGGLE_ALL_DEVICES_DESIGN_MODE,
+    allDevicesInDesignMode,
+  };
+}
+
+export function toggleDeviceDesignMode(deviceId, designModeOn) {
```
sidthesloth92, line 198: Do we really need a separate dispatch function? Can we keep this and remove the dispatch?

desktop-app/flow-typed/electron.js
```diff
@@ -1,6 +1,6 @@
 declare class WebviewElement extends HTMLElement {
```
sidthesloth92, line 1: Is this intention. Also, since this is a `.js` do we need to add flow before using typings?

desktop-app/app/actions/browser.js
```diff
@@ -698,6 +716,23 @@ export function onDeviceMutedChange(deviceId, isMuted) {
...
 }
 
+export function onAllDevicesDesignModeChange() {
+  return (dispatch: Dispatch, getState: RootStateType) => {
+    const {
+      browser: {allDevicesInDesignMode},
+    } = getState();
+    const next = !allDevicesInDesignMode;
+    pubsub.publish(TOGGLE_DEVICE_DESIGN_MODE_STATE, [{designMode: next}]);
+    dispatch(toggleAllDevicesDesignMode(next));
```
sidthesloth92, line 726: Since it's a toggle, we don't need to pass a value as we retrieve the previous value from the state in the reducer.

desktop-app/app/reducers/browser.js
```diff
@@ -517,6 +521,28 @@ export default function browser(
           proxy: action.profile,
         },
       };
+    case TOGGLE_ALL_DEVICES_DESIGN_MODE:
+      const nextDevices = state.devices;
```
sidthesloth92, line 525: I believe we shouldn't be modifying state here. Also, since it's toggle, can we rely on the previous state instead of an action value? Can we do like,

```
const nextDevices = [...state.devices];
nextDevices.forEach(d => (d.designMode = !state.allDevicesInDesignMode));
 return {
        ...state,
        allDevicesInDesignMode: !state.allDevicesInDesignMode,
        devices: nextDevices
};
```

desktop-app/app/reducers/browser.js
```diff
@@ -517,6 +521,28 @@ export default function browser(
...
+      nextDevices.forEach(d => (d.designMode = action.allDevicesInDesignMode));
+      return {
+        ...state,
+        allDevicesInDesignMode: action.allDevicesInDesignMode,
+        devices: nextDevices,
+      };
+    case TOGGLE_DEVICE_DESIGN_MODE:
+      const deviceIndex = state.devices.findIndex(
```
sidthesloth92, line 533: I think we should do, `nextDevices = [state.devices]` to avoid modifying the state.

desktop-app/app/reducers/browser.js
```diff
@@ -332,6 +335,7 @@ export default function browser(
     windowSize: getWindowSize(),
     allDevicesMuted: false,
     networkConfiguration: _getNetworkConfiguration(),
+    allDevicesInDesignMode: false,
```
sidthesloth92, line 338: I think we should also set individual device status to `false` at end of `_getActiveDevices` function.

desktop-app/app/components/WebView/index.js
```diff
@@ -426,6 +441,14 @@ class WebView extends Component {
     this.getWebContents().setAudioMuted(muted);
   };
 
+  processToggleDesignModeEvent = ({designMode}) => {
+    this.webviewRef.current
+      .executeJavaScript(
+        `document.designMode = "${designMode ? 'on' : 'off'}";`
```
sidthesloth92, line 447: Can we externalize on and of as `enums`?

sidthesloth92 (changes requested):

desktop-app/app/actions/browser.js
jjavierdguezas: I just followed "the style" of the code around, but I think is ok to change it

desktop-app/app/actions/browser.js
jjavierdguezas: I think it's ok to change it

desktop-app/app/actions/browser.js
jjavierdguezas: that makes sense

desktop-app/app/reducers/browser.js
jjavierdguezas: I think it will be _falsy_ by default, but yes is better to set it `false`

desktop-app/app/reducers/browser.js
jjavierdguezas: >rely on the previous state instead of an action value

In order to keep in sync with other comments, I think it is ok to change it

>I believe we shouldn't be modifying state here

doing `const y = [...x]` only copies the array, the items inside `y` are the same (same reference) as in `x`. to do it right you need to deep clone the elements too, and I don't think it is worth it

desktop-app/app/reducers/browser.js
jjavierdguezas: see my previous comment

desktop-app/flow-typed/electron.js
jjavierdguezas: I change this in order to pass `eslint` here in GitHub Action

manojVivek:
@sidthesloth92 I'm merging this in. If you have any follow up questions, please comment those and can be addressed in a new PR.

sidthesloth92:
Sure @manojVivek

desktop-app/app/reducers/browser.js
sidthesloth92: Understood.

PR #517, opened 2020-10-16
[Feature] device limit warning

### LINKED ISSUE #420: Show a warning in the Device manager when more number of devices(more than what the system resources can handle) are added for preview
### Problem
Users adding more number of devices to the preview than their system can handle.

### Solution
1. Come up with a device per CPU value for optimal app functioning.
2. When the user adds more than what their system can handle, show a warning `Adding more than "x" devices may slow down the system. Are you sure you want to continue?`.

jjavierdguezas:
# ✨ Pull Request

Resolves #420 

### ℹ️ About the PR

We came out with this idea to calculate a recommended max number of devices based on machine capabilities. The idea is based on logical CPUs amount and their speed. The final version is based on observations on our own machines, but we need to review it in a while and tweak it as needed

wrn


manojVivek (approved):
Looks great!

PR #522, opened 2020-10-24
feat(desktop-app/app): show url link on hover

### LINKED ISSUE #516: Show link url on hover
# 🚀 Feature Request

### 📝 Description
In standard browsers, hovering over a link triggers the display of the full url at the bottom left of the screen.
This is not replicated in responsively.

Same behaviour as standard browsers, ie display link url

rishichawda:
# ✨ Pull Request

### 📓 Referenced Issue

resolves #516 

### ℹ️ About the PR

The URL from the `href` property for the anchor element is displayed in the bottom left of the screen on hover.

Notes: 
1. Added listener to document element for `mouseout` event as it gets sent to child elements. With `mouseover`, we would need to target individual links on all webviews - which is not feasible.
2. Kept the position and design similar to general browsers.
3. This is my first time contributing to this project, so if I missed something - please point me in the right direction so I can better understand and improve the code to best fit the existing patterns or into the codebase as a whole. 😄 

Screenshot 2020-10-24 at 5 20 20 PM

Screenshot 2020-10-24 at 6 02 43 PM


jjavierdguezas:
Hi @rishichawda thanks for this contribution!
I'll review it in a bit
How does it look like when Light theme is on?

rishichawda:
@jjavierdguezas It would look the same irrespective of the theme.

jjavierdguezas (changes requested):
Looks very good @rishichawda !!
Just two requests:
- Please do not mix `makeStyles` function and `style.css` file. Use the first and remove the file (and use
```diff
- import {makeStyles} from '@material-ui/core';
+ import {makeStyles} from '@material-ui/core/styles';
```
in order to use `theme.palette.mode` )

- I think it is too dark for light theme (@manojVivek @jzabala what do you think?)
using this for `.container`:
```js
backgroundColor: theme.palette.mode({
      light: '#e0e0e0',
      dark: '#424242',
    }),
```
and this for `.text`
```js
color: theme.palette.mode({
      light: '#000',
      dark: '#fff',
    }),
```
I think the result is better for the theme
image

What do you think? @rishichawda

rishichawda:
Yeah, even I thought of doing a light theme in the morning today but held it off until I heard back from you on this PR. Your design looks cool too. I thought of doing something similar where the container's background matched one of the existing greys from the application's palette on light mode. If everyone's okay with it, I can make the changes and push it tomorrow.  🤟 😄

manojVivek:
> I thought of doing something similar where the container's background matched one of the existing greys from the application's palette on light mode

@rishichawda That would be great. Also, you don't have to set different colors in the styles based on the selected theme. Just go with a generic `theme.palette.background.l*` like in this one: https://github.com/responsively-org/responsively-app/blob/c970238493a4e41e7f505f09a8fdaebe4f523b2f/desktop-app/app/components/AddressInput.js#L323

So that the right color is applied automatically based on the selected theme instead of handpicking it.

And for the text color, you can go with `theme.palette.text.normal`.

desktop-app/app/containers/Browser/index.js
```diff
@@ -63,6 +64,7 @@ const Browser = ({browser}) => (
       </div>
     </div>
     <AppNotification />
+    <LinkHoverDisplayContainer />
```
manojVivek, line 67: @rishichawda Can you please move this to the `DevicesPreviewer` component so that this info just stays within the page preview area? 

97111563-ca26f180-16df-11eb-928e-b672640f0eda

rishichawda:
@manojVivek I'll update the PR soon. Thanks! 😄

desktop-app/app/containers/Browser/index.js
rishichawda: Yup this seems to be a better way. Thanks, I'll update this. 👍

manojVivek:
@jjavierdguezas Looks good, merging it. 

@rishichawda Kudos to your work on implementing this feature! 🎉

PR #533, opened 2020-11-11
Option to set blank page on application launch (#521)

### LINKED ISSUE #521: Option to set blank page on application launch
# 🚀 Feature Request

### 📝 Description

On launching the application all the viewpoints are refreshed and load Google.com. Please add an option to have it load a blank page so we don't have to wait for them all to load before using the application.

Blank page on startup checkbox in the settings menu.

bikingbadger:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes #521 

### ℹ️ About the PR

Added a dropdown list with two options for either a homepage or blank. Depending on this setting when the app starts the preference selected will show the relevant page


bikingbadger:
This is the first time actually delving into someone else's codebase and actually making changes for a PR so I would really appreciate any feedback on how I could improve on what I did and maybe anything to look out for. This is also the first time looking at a React app so coming from Vue I was able to make out what should be done but I'm again here to try and learn and help where I can. Thanks in advance.

desktop-app/app/components/UserPreferences/index.js
```diff
@@ -245,6 +247,29 @@ function UserPreference({
...
+        </div>
+        <div className={classes.marginTop}>
+          <Select
+            options={startOptions}
+            value={
+              startOptions.find(x => x.value === userPreferences?.startPage) ||
+              startOptions[0]
+            }
+            onChange={val => {
+              notifyStartPreferenceChanged(val.value);
```
manojVivek, line 267: @bikingbadger Where is this notification consumed?

manojVivek:
> This is the first time actually delving into someone else's codebase and actually making changes for a PR so I would really appreciate any feedback on how I could improve on what I did and maybe anything to look out for. This is also the first time looking at a React app so coming from Vue I was able to make out what should be done but I'm again here to try and learn and help where I can. Thanks in advance.

@bikingbadger This looks great. 👏🏼
Can you please attach a screenshot of the user preferences section after this change?

desktop-app/app/components/UserPreferences/index.js
bikingbadger: Currently, it isn't and is something that probably needs discussion. When changed should it update the current page? Or only on the next start? This part could then be removed if it is only done on startup.

bikingbadger:
Screenshot 2020-11-12 091125

Here is where I placed it.

desktop-app/app/components/UserPreferences/index.js
jjavierdguezas: I think I shouldn't refresh the current page, it would take effect on the next startup

desktop-app/app/utils/navigatorUtils.js
```diff
@@ -13,7 +15,10 @@ export function saveLastOpenedAddress(url) {
 }
 
 export function getHomepage() {
-  return settings.get(HOME_PAGE) || 'https://www.google.com/';
+  return userPreferenceSettings.getDefaultStartPage() ===
+    START_MANAGEMENT_OPTIONS.HOME
+    ? settings.get(HOME_PAGE) || 'https://www.google.com/'
+    : 'about:blank';
 }
```
jjavierdguezas, line 22: I think this function is used in more than one place, for example, what exactly happens in this scenario:
1- Open the app
2- Go to https://responsively.app
3- Set https://responsively.app as the home page 
4- Set the blank page as the startup page, 
5- Close the app
6- Open the app (the blank page should be opened)
7- Go to any website
8- Click on Go to Home Page button
9- **Does (should) it go to the black page or to responsively website ? 🤔**

jjavierdguezas:
Should we put some code in [migration.js](https://github.com/responsively-org/responsively-app/blob/master/desktop-app/app/settings/migration.js) to ensure a default value in the user settings?
@manojVivek

desktop-app/app/utils/navigatorUtils.js
bikingbadger: I thought that this was used only on startup. So I think you may be right that it should go to Home that was set as home and not the blank. That is at least what the expected result would be if I was looking at it. I will have to take a look at home this is used in the startup to use it there. Or do you think that this is an acceptable solution?

manojVivek:
> Should we put some code in [migration.js](https://github.com/responsively-org/responsively-app/blob/master/desktop-app/app/settings/migration.js) to ensure a default value in the user settings?
> @manojVivek

Yeah. I would suggest keeping `START_MANAGEMENT_OPTIONS.HOME` as the default value for the startup page, so the current behavior is restained and those who need can change it from the preferences. 
cc: @bikingbadger

desktop-app/app/utils/navigatorUtils.js
jjavierdguezas: That's what I think. The behavior after step 8 should be to go to the configured home page (https://responsively.app), not the startup page (_blank)

bikingbadger:
Could you explain a bit more on migration.js I've gotten a bit lost in how you want this to be handled

manojVivek:
> Could you explain a bit more on migration.js I've gotten a bit lost in how you want this to be handled

@bikingbadger This is for setting the default value for this user preference when the app starts after this update. 

In this file, https://github.com/responsively-org/responsively-app/blob/master/desktop-app/app/settings/migration.js, we handle the defaults when the app starts.

manojVivek:
> @manojVivek So this setting should be saved to there?

Yes, this setting should be saved with the default value as "Homepage" in the migration.js file, as that is the current behaviour.
> 
> As for the issue of pressing the Home button what will be the expected behavior if blank is set

The home button press should take them to the saved homepage URL. This setting is just to control what should be loaded when the app is opened.

jjavierdguezas:
Hi @bikingbadger , are you still working on this? just checking....

jjavierdguezas:
I'm closing this old PR and I'm going to work on this issue

PR #538, opened 2020-12-01
Save Layout settings

### LINKED ISSUE #534: Save Layout settings
# 🚀 Feature Request

### 📝 Description

I always want a `Horizontal` layout but it always switches back to `FlexiGrid`

Save whatever layout I select before exiting the app & show me the same layout when I open the app.

At least, that's what Desktop Apps are supposed to do :)

kiwan97:
# ✨ Pull Request

### 📓 Referenced Issue

Resolves #534 

### ℹ️ About the PR

I edited ./desktop-app/app/constants/settingKeys.js and ./desktop-app/app/reducers/browser.js
This is my second PR in my life!
I'm not sure this is right.

if I changed Layout setting to "HORIZONTAL",
I could find that Layout setting is still "HORIZONTAL".
I also tested two remainings.

### DIFF
--- a/desktop-app/app/constants/settingKeys.js
+++ b/desktop-app/app/constants/settingKeys.js
@@ -5,3 +5,4 @@ export const NETWORK_CONFIGURATION = 'networkConfiguration';
 export const BOOKMARKS = 'bookmarks';
 export const STATUS_BAR_VISIBILITY = 'statusBarVisibility';
 export const APP_NOTIFICATION = 'appNotification';
+export const LAYOUT = 'layout';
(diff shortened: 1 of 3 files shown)


jjavierdguezas (changes requested):
Hi @kiwan97 this looks great
Just one request, please add code to the [migration.js](https://github.com/responsively-org/responsively-app/blob/master/desktop-app/app/settings/migration.js) file to ensure a default value in the user settings
It should be easy just looking around in that file (for example the `_handleColorThemePreferences` function)

PR #546, opened 2020-12-30
Update brew cask command

karbassi:
Brew has updated to only allow `brew install --cask`.

```sh
Error: Calling brew cask install is disabled! Use brew install [--cask] instead.
```


jjavierdguezas:
thanks for this fix

PR #578, opened 2021-05-08
Add screen shot all devices functionality

### LINKED ISSUE #541: Ability of take a screenshot of all the devices at once
# 🚀 Feature Request

### 📝 Description
I tend to share screenshots of all the devices at once with my team. Currently I just use a screenshot tool but it would be nice to be able to have it done by the app. As example
image

The app could create a screenshot of it's content and scale it.

Just using the screenshots app from my OS

durgakiran:
# ✨ Pull Request

### 📓 Referenced Issue
closes #541 

### ℹ️ About the PR
Give take a screenshot of all device Web views at once, including merging of images.

Features:
<h2>

```diff
+ Choosing screens to capture
```
</h2>

image

<h2>

```diff
+ Capture fail and recovery mechanism
```
</h2>

One of the existing issues with the existing screenshot mechanism is it extremely fail prone. The main reason as I observed is crash electron main render thread, which fails and never recovers when continuously `scrollViewPort` function is called. One more issue with the current screenshot capture is if the render thread fails to recover entire application needs to be reloaded.

To counter such fails we introduce a function that returns a promise with a retry mechanism limited to 3 retries. The function listens to webContents `crashed` event. we use `crashed` event to understand the if webview is killed or just crashed. if Webview is just crashed, the code will again load the crashed web view. Once webview reloads function will listen to `dom-ready` event. which notifies once the web view is completely loaded, then we try to capture the image again.

```
captureWithRetry(retry: number = 3): Promise<Electron.NativeImage> {
    return new Promise((resolve, reject) => {
      let retryCount = 0;
      const address = this.webView.getAttribute('src');

      const handler = (event, killed) => {
        if (!killed) {
          this.webView.setAttribute('src', address);
        } else {
          reject(new Error('WebView killed.... : ('));
        }
      };
      this.webContents.on('crashed', handler);

      const callGetImage = () => {
        // noinspection JSIgnoredPromiseFromCall
        getImage();
      };

      this.webContents.on('dom-ready', callGetImage);

      const getImage = async () => {
        if (retryCount === retry) {
          reject(new Error('Unable to produce image'));
        }
        retryCount += 1;
        const image = await this.captureFullHeightScreenShot();
        this.webContents.removeListener('dom-ready', callGetImage);
        this.webContents.removeListener('crashed', handler);
        resolve(image);
      };

      // noinspection JSIgnoredPromiseFromCall
      getImage();
    });
  }
```
If Image capture fails beyond the threshold i.e, 3. The code will skip that particular web view and proceed with other web views.
This way we are not abandoning the app in a failed state.

<h2>

```diff
+ Merging images
```
</h2>
Each web view, whose image is captured, binary data of the image is store in the memory. Once completion of taking individual screenshots is done. The binary data of each data is combined by drawing on a canvas. The canvas data again converted to binary and saved to the folder. The merging of images **does not** block Electron's main threads.

<h2>

```diff
! Caution: Merging of images takes a lot of time
```
</h2>

<h2>

```diff
+ Final indication
```
</h2>
Once the capture and saving process is over, automatically the created folder is opened.

**There are some improvements in code refactoring and structure, I think those are self-evident in the PR file changes.**

[please click here to view feature demo video](https://www.loom.com/share/70cb2f3df14d474683d2f4170c5708e2).

Example merged screenshots.
Google-all-2021-10-5


manojVivek:
@durgakiran I think, it would be more useful if we add the device name above the screenshot. What do you think? 

This will help the use case of showcasing the screenshots.

durgakiran:
> @durgakiran I think, it would be more useful if we add the device name above the screenshot. What do you think?
> 
> This will help the use case of showcasing the screenshots.

I thought about it and forgot to add that functionality. I will update the PR. any other changes you recommend?

durgakiran:
I am converting this pull request into draft as I observed a bug.

durgakiran:
@manojVivek   one thing I observed was when a screenshot of a web view is happening, the screenshot of that webview is not happening until that web view is in the view port of the electron, any reason in your mind about this behaviour? It just does not seem logical to me? what do you think?

manojVivek:
> @manojVivek one thing I observed was when a screenshot of a web view is happening, the screenshot of that webview is not happening until that web view is in the view port of the electron, any reason in your mind about this behaviour? It just does not seem logical to me? what do you think?

@durgakiran Yeah that is how Chromium behaves. Maybe they don't render/paint the sections of the screen if they are outside the viewport. 
We handle this behaviour by zoom-out the web-views before the screenshot is taken so that all the web-views are in the viewport and then process the screenshot. Did you face it during your testing? Ideally, you shouldn't have.

durgakiran:
@manojVivek yes, yes I faced it. Maybe because I was almost completely recoded the screenshot all functionality. So, I might have missed something.

durgakiran:
Ok I will look into that, thanks

desktop-app/app/components/ScreenshotManager/Capture.js
```diff
@@ -0,0 +1,268 @@
...
+  const images = [];
+  // device names
+  const deviceNames = [];
+
+  // get web address of Web views
+  const address = sortedWebViews[0].getURL();
+
+  const fsUtils = new FileSystemUtils(path, defaultPath, now, address);
+
+  for (let i = 0; i < sortedWebViews.length; i += 1) {
```
manojVivek, line 89: @durgakiran Processing the screenshots sequentially is going to take a very long time. We should look at parallelizing the capture here.

desktop-app/app/components/ScreenshotManager/Capture.js
```diff
@@ -0,0 +1,268 @@
...
+  // open directory
+  await fsUtils.openCurrentDir();
+}
+
+// noinspection JSUndefinedPropertyAssignment
+export async function mergeImages(
+  images: Array<Electron.NativeImage>,
+  deviceNames: Array<string>
+) {
+  const canvas = document.createElement('canvas');
```
manojVivek, line 185: @durgakiran Is there a better way to do the merging instead of using a `canvas` in terms of performance? 
Just to make sure the screenshot functionality finishes as quickly as possible.

manojVivek (commented):
@durgakiran Great effort in getting this to PR. 👏🏼 
 We just need to tweak the design of the merged image a bit as well so that it looks more presentable.

Maybe a `lightgray` background, `black` color for the devices' names and some margin around the final image would make it look better?

desktop-app/app/components/ScreenshotManager/Capture.js
durgakiran: As far as I observed, we can't use node workers in electron. I think parallelization in Electron can only be achieved through web workers but web workers don't have access to webviews. So I think we kind of hit a big roadblock there. Do you have any suggestions?

desktop-app/app/components/ScreenshotManager/index.js
```diff
@@ -1,5 +1,122 @@
...
+        ...device,
+        selected: !device.selected,
+      };
+    });
+    setDeviceChecks(newState);
+  };
+
+  const handleGo = () => {
+    const payload = {
+      zip,
```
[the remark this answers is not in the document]
durgakiran: I forgot to remove `zip`. I will update that

desktop-app/app/components/ScreenshotManager/Capture.js
durgakiran: @manojVivek I think, we should really explore here. So far I did not find any, but I believe there can be better methods. maybe we should explore a bit of web assembly.

durgakiran: I am exploring `webGL`, seems promising. I will update you if I find a real good way.

desktop-app/app/components/ScreenshotManager/Capture.js
durgakiran: I could not find any alternative here

desktop-app/app/components/ScreenshotManager/Capture.js
durgakiran: I have optimized the process a bit, the merging and saving is fast now.

durgakiran (commented):
As far as design changes for png background, I could not see any point in it. It unnecessarily increases the file size.

manojVivek:
> As far as design changes for png background, I could not see any point in it. It unnecessarily increases the file size.

We can save the image as `jpeg` and that will have reduced file size. 
The reason why I'm insisting it is having non-transparent background is that is going to make it easier for the users to just share/use it without any additional processing.

desktop-app/app/components/ScreenshotManager/Capture.js
manojVivek: @durgakiran Why don't you make it an `async.map` or something like that? 
So that the capture can happen in parallel whenever the CPU is free. We have some intentional `sleeps` in the capture workflow and this parallelization will cut down some waiting time over that.

rille111:
Hey still is in PR.. would appreciate this functionality. :)

durgakiran:
@manojVivek  thought of picking this up again but found a lot has been changed, since this PR. Will work on this PR again

manojVivek:
@durgakiran Great to see you back again. 
Yeah the app has went through a full revamp, and should be relatively easier to add new features.

Issue #365, opened 2020-07-30
🐛: When using 'open with' on HTML files it loads google.com page instead of the HTML file.


Kidcredo:
![IwAKSPmTRP](https://user-images.githubusercontent.com/15522605/88883541-21720e80-d224-11ea-8883-51c6cabb05f6.gif)

manojVivek:
That is strange.
It tries to loads that file and then goes back to google.com. 

@esprush Could this be because of the URL syncing? 

@jjavierdguezas Any ideas?

jjavierdguezas:
I cant reproduce it, @Kidcredo can you please share the files?

jjavierdguezas:
Thats weird @Kidcredo , it works in my PC like a charm

I completed the `fun.js` Here you have the used files: https://we.tl/t-rCG0FduozF

![openwithissue2](https://user-images.githubusercontent.com/13673443/88938316-8b6ed000-d285-11ea-904c-145b5b95aa24.gif)

Kidcredo:
I am guessing this was because of some cache issue due to updating directly without uninstalling the 0.8.0 version first. 
Just uninstalled and did a clean install now it seems to be working fine.

Kidcredo:
After the clean install it worked once and when I tried it again it happened again.
It happened again with the file you shared @jjavierdguezas 
![9DtPEXVKBx](https://user-images.githubusercontent.com/15522605/88943673-4d6cae80-d27b-11ea-9f98-558d0700c745.gif)

jjavierdguezas:
Definitely something weird happens. I can see by myself that when you open a file using `Open with`, sometimes Responsively go a back to HOME_PAGE and the file, its like some race condition between redux state initialization and the open file logic. Somehow in my tests it always ends showing the desired file, but in your PC "got to home page" wins...

jjavierdguezas:
I tried with multiple devices and in my PC it always works,

manojVivek:
@jjavierdguezas @Kidcredo It is possible it could be because of the URL sync.

Possible explanation:
1. One of the devices in the preview is lagging in rendering during the app load.
2. When the `address-change` event is triggered. It is failed in the lagging device because the `webview` is not ready.
3. Finally, when the lagging device loads the default URL(google.com), the URL sync is transmitting that URL to all other devices.
4. Other devices process the event and load google.com as well.

If we watch the address bar URL changes closes, it confirms this behavior. 

cc: @esprush

[...]

PR #343, opened 2020-07-23
Ux addressbar

abakermi:
UX adressbar  , `KeyDown` ,`KeyDown` , `Escape` #338

PR #586, opened 2021-05-12
Allow pass url/filepath from cli args

### LINKED ISSUE #369: Open the App from the cli and load the url/filepath passed as the arg.
Sample Mac command can be:
```
open /Applications/ResponsivelyApp.app --args https://example.com
```

jjavierdguezas:
# ✨ Pull Request

### 📓 Referenced Issue

resolves #369

and maybe, just _maybe_ #365 

### ℹ️ About the PR

Set `process.argv[1]` as `urlToOpen` if no url or file has been set to open.
As part of this PR I had to change the place in the code where the initial URL is set and as a consequence it is _possible_ that #343 has been resolved (at least an improvement should be appreciated) 

I tested this on Windows only, it should be tested on other OS too


manojVivek (approved):
Works great on Mac as well. Merging it.

PR #588, opened 2021-05-17
Add option to switch between blank page and homepage when app launches

### LINKED ISSUE #521: Option to set blank page on application launch

jjavierdguezas:
# ✨ Pull Request

### 📓 Referenced Issue

resolves #521 

### ℹ️ About the PR

Added a select to User Preferences to set between blank page and homepage as startup page when app launches

image

❗ An scrollbar appeared, maybe it's time to reorder the left menu 👀
I tried with a checkbox like "Use blank page instead of homepage when app launches" but it take two lines and the scrollbar appeared too

image

and another version:

image

I think we could go like this and then reorder the whole sidebar

### DIFF
--- a/desktop-app/app/actions/browser.js
+++ b/desktop-app/app/actions/browser.js
@@ -52,6 +52,7 @@ export const TOGGLE_DEVICE_DESIGN_MODE = 'TOGGLE_DEVICE_DESIGN_MODE';
 export const SET_HEADER_VISIBILITY = 'SET_HEADER_VISIBILITY';
 export const SET_LEFT_PANE_VISIBILITY = 'SET_LEFT_PANE_VISIBILITY';
 export const SET_HOVERED_LINK = 'SET_HOVERED_LINK';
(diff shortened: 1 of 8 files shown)


manojVivek:
@jjavierdguezas The 3rd variation looks good but that still introduces the scroll bar right? 

Also, it depends on the user screen size anyways, users with a smaller monitor are still going to see the scrollbar.

jjavierdguezas:
> @jjavierdguezas The 3rd variation looks good but that still introduces the scroll bar right?
> 
> Also, it depends on the user screen size anyways, users with a smaller monitor are still going to see the scrollbar.

yes, all versions introduce scroll at least in resolutions like 1920x1080. I think we can merge this and reorder the left menu in another issue

manojVivek (approved):
Sounds good, merging now.

Issue #530, opened 2020-11-04
When devtools is right-docked, items above cannot be clicked


alanw83:
# 🐞 (Original) bug report

### ✍️ Description

If I try and open the Live CSS editor when devtools is open it won't all me to open the live css editor. I have to open live css editor then dev tools to have them both open at the same time.

### 🕵🏼‍♂️ Is this a regression?

Live CSS editor was not in previous version

### 🔬 Minimal Reproduction

Open program
Load website
Toggle Live CSS Editor to on
Inspect an element to open dev tools
Attempt to toggle off the Live CSS editor - Hand to click does not appear

### 🔥 Exception or Error or Screenshot

<pre><code>

</code></pre>

see: https://github.com/responsively-org/responsively-app/issues/530#issuecomment-722350840

jjavierdguezas:
Hi @alanw83 , I couldn't reproduce it.
This is my environment
```
Version: 0.15.0
Electron: 9.3.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.18362
```

Could you please share your environment? (Press F1 on the app then click Copy button an paste here)
Could you also share a gif or a video showing the issue?

alanw83:
Version: 0.15.0
Electron: 9.3.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.19041

https://drive.google.com/file/d/1xJvX8Gglxb2y_ybWFXqkXrm_TIbwuSlv/view

Hopefully this works and helps!

jjavierdguezas:
Thanks for the info @alanw83 

I keep the devtools docked to the bottom, that is why I couldn't reproduce it. Now I did,

I think you found a more complex issue: **when the devtools is docked to the right side of the app, it seems that it is positioned on top of the elements that are above, preventing them from being clicked**

I mean, not only does the Live CSS Editor toggle button stop working, but all buttons stop working as well. Also, if you make the devtools wider, you won't be able to click on the right side of the navigation bar or the action buttons found there.

cc: @manojVivek 

![devtools issue](https://user-images.githubusercontent.com/13673443/98241777-bdba5880-1f6b-11eb-8251-2e99e77cef21.gif)

Zoeinfp:
I was unable to reproduce following those steps because the live css does toggle off ...

Version: 0.15.0
[...]

PR #662, opened 2022-01-24
fix bug #530 : items above devtools cannot be clicked

zakariazuck:
# ✨ Pull Request

### 📓 Referenced Issue

fix bug #530

### ℹ️ About the PR

Add z-index to the style of the Header with value bigger than the one of the devtools


manojVivek:
@zakariazuck I don't think this is the right fix as well. The issue is not when we have the shortcuts window open, it is when we have the dev-tools open and docked to the right.

LolaBanny (approved):

PR #706, opened 2022-06-30
define a variable on the window object to help analytics

### LINKED ISSUE #696: Detect responsively.app so we can tag it as robot traffic
# 🚀 Feature Request

### 📝 Description

We run a analytics company, Simple Analytics, and some customer sees traffic of many browsers when using Responsively.app. This makes sense, but for us there is no way to detect the traffic because the user agents seem to not include the Responsively.app name.

Some tools define a variable on the window object. For example `window._bot = true`. That would be super helpful.

N/A

jjavierdguezas:
# ✨ Pull Request

### 📓 Referenced Issue
resolves #696

### ℹ️ About the PR

set `window._bot = true` on each device

image

### DIFF
--- a/desktop-app/app/components/WebView/index.js
+++ b/desktop-app/app/components/WebView/index.js
@@ -185,6 +185,13 @@ class WebView extends Component {
     );
 
     this.webviewRef.current.addEventListener('dom-ready', () => {
+      this.getWebContentForId(this.webviewRef.current.getWebContentsId())
+        .executeJavaScript(
+          `{
+            window._bot = true;
+          }`
+        )
+        .catch(captureOnSentry);
       this.initEventTriggers(this.webviewRef.current);
       this.dbg = this.getWebContents().debugger;
       if (!this.dbg.isAttached()) {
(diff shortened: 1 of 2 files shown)


desktop-app/app/components/WebView/index.js
```diff
@@ -185,6 +185,13 @@ class WebView extends Component {
     );
 
     this.webviewRef.current.addEventListener('dom-ready', () => {
+      this.getWebContentForId(this.webviewRef.current.getWebContentsId())
+        .executeJavaScript(
+          `{
+            window._bot = true;
```
manojVivek, line 191: This runs after `dom-ready`, what if the analytics script executed before this? 🤔

jjavierdguezas: If I move it outside `dom-ready` event (i.e.: the start of `componentDidMount`) I get:
image

So, I moved the `window._bot = true` line to `preload.js` and it seems to work

what do you think @manojVivek?

manojVivek: Yes, this looks good JJ!

manojVivek (approved):

w4rner:
when will this merge go live into the product?

PR #856, opened 2023-03-14
Update README.md

Archit0308:
Elaborated the README.md file.

Explained the features in a more understandable manner. Ps: Just wanted to do a little meaningful contribution ^_^

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@
 5. One-click screenshot all your devices.
 6. Hot reloading supported for developers.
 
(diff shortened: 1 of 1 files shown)


codewithmitesh (approved):

manojVivek:
To me, it looks like a repetition of what's said just above this, feels redundant. Also, this makes the README so content heavy, while I think we should strike to keep this simpler.

I'm closing this with the above note, please let me know if you have concerns.

Issue #812, opened 2023-02-17
Add shortcut for opening dev tools


manojVivek:
### Discussed in https://github.com/responsively-org/responsively-app/discussions/811

<div type='discussions-op-text'>

<sup>Originally posted by **dev-song** February 17, 2023</sup>
I'm using macOS Ventura 13.1... (Intel Mac)
Is there a shortcut for `Inspect Element` or opening a dev tool?
I'd be much better if I can check elements without moving mouse cursor to top-right of the app everytime...</div>

ahmarammar:
Hey, @manojVivek I hope you are doing well, I think you are looking for shortcuts that help you to open developer tools in your macOS.
1 - Command+Option+C  (It will open the Developer Tools and highlight the element you clicked on)
2 - Command+Option+I (It will open the Developer Tools)

Ayan-Kumar-Saha:
Hi @manojVivek Yes there is a shortcut. Use `Command + Option + I` to open the dev tool.

BrijenMakwana:
Can you please assign this to me? I would love to work on it.

EvokerKing:
Quick suggestion for this: do something like Polypane does where if you hold down alt, it switches to inspect element mode and switches back when you release alt.

PR #873, opened 2023-03-23
feat!: add shortcut for opening devtools

codewithmitesh:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes: #812 

### ℹ️ About the PR

Included a functionality that allows users to press the "S" key to select the inspect element instead of having to move the mouse cursor to the top-right corner of the application every time. Pressing "S" key again will deselect the inspect element

<pre>
<a href="https://gifyu.com/image/SIv2a"></a>
</pre>


desktop-app/src/renderer/components/ToolBar/index.tsx
```diff
@@ -6,17 +6,39 @@ import {
...
 import Menu from './Menu';
 import Button from '../Button';
 import AddressBar from './AddressBar';
 import ColorSchemeToggle from './ColorSchemeToggle';
 
+function useKey(key, cb) {
+  const callbackRef = useRef(cb);
+  useEffect(() => {
+    callbackRef.current = cb;
+  });
```
manojVivek, line 20: Can you please add `[cb]` as a dependency to this effect? As we don't want this to run on every render.

desktop-app/src/renderer/components/ToolBar/index.tsx
```diff
@@ -6,17 +6,39 @@ import {
...
+}
 const ToolBar = () => {
   const rotateDevice = useSelector(selectRotate);
   const isInspecting = useSelector(selectIsInspecting);
   const dispatch = useDispatch();
 
+  function handleInspectShortcut() {
+    dispatch(setIsInspecting(!isInspecting));
+  }
+  useKey('KeyS', handleInspectShortcut);
```
manojVivek, line 40: Can you please validate once that this listener to `s` key is not hijacking user inputs to text boxes? 

On both the address bar and some input boxes in the previewed page as well.

manojVivek (changes requested):
Just a couple of minor comments, looks 🔥🔥 otherwise!!!

desktop-app/src/renderer/components/ToolBar/index.tsx
manojVivek: Also, is there a specific reason you chose `s` as the shortcut? 

I would prefer `i` for `inspect element`.

desktop-app/src/renderer/components/ToolBar/index.tsx
```diff
@@ -6,17 +6,40 @@ import {
...
 } from 'renderer/store/features/renderer';
 import { Icon } from '@iconify/react';
+import { useEffect, useRef } from 'react';
 import NavigationControls from './NavigationControls';
 import Menu from './Menu';
 import Button from '../Button';
 import AddressBar from './AddressBar';
 import ColorSchemeToggle from './ColorSchemeToggle';
 
+function useKey(key: any, cb: any) {
```
manojVivek, line 16: Use of `any` should be avoided as much as possible as it beats he purpose of ts. Any specific reason why you chose to make these `any`?

This can be `function useKey(key: string, cb: () => void)`

codewithmitesh: No, There is no specific reason, I am just new with typescript😅.

manojVivek: @codewithmitesh I'm planning to cut a release this week, can you resolve this so that we can merge this in for the release.

manojVivek (approved):
This is looking great, looking forward to more like this!

manojVivek:
@all-contributors Please add @codewithmitesh for code.

PR #914, opened 2023-05-05
Ability to Rotate Only One Device, Not All at Once

WayneRocha:
# ✨ Pull Request

### ℹ️ About the PR

The main rotation button are with the same behavior. But each device has a button that control their own status and when activated, lock the interaction with the main rotation button.

All the screens were controlled by one boolean variable in redux store. Now that variable is an object with the [following interface](https://github.com/WayneRocha/responsively-app/blob/main/desktop-app/src/renderer/store/features/renderer/index.ts#L9). It has one boolean property for main button, and other for each individual device status, keyed by the _name_.

image
image
44-8893-4b84-b8ef-7f2c5d277f18.png)

### DIFF
--- a/desktop-app/src/renderer/store/features/renderer/index.ts
+++ b/desktop-app/src/renderer/store/features/renderer/index.ts
@@ -3,10 +3,12 @@ import type { PayloadAction } from '@reduxjs/toolkit';
 import { PreviewLayout } from 'common/constants';
 import type { RootState } from '../..';
 
+type IRotatedDevices = { [key: string]: { inSingle: boolean, rotate: boolean } };
+
 export interface RendererState {
   address: string;
   zoomFactor: number;
-  rotate: boolean;
+  rotatedDevices: IRotatedDevices;
   isInspecting: boolean | undefined;
   layout: PreviewLayout;
 }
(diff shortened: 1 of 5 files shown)


desktop-app/src/renderer/store/features/renderer/index.ts
```diff
@@ -27,7 +32,7 @@ const urlFromQueryParam = () => {
 const initialState: RendererState = {
   address: urlFromQueryParam() ?? window.electron.store.get('homepage'),
   zoomFactor: zoomSteps[window.electron.store.get('renderer.zoomStepIndex')],
-  rotate: false,
+  rotatedDevices: { allRotated: false, devices: {} },
```
manojVivek, line 36: @WayneRocha Just speaking my thoughts here, I kind of feel like this implementation with `rotatedDevices` is a little more complicated than this functionality needs to be. We're relying too much on the Redux state here while I have a possible solution without that, let me know your thoughts on the below approach. 

Instead of adding `rotatedDevices` state in the redux store to maintain the rotation, we can retain the implementation of the main toolbar rotate button as such with a boolean field in the redux store and for the individual devices we can maintain a local state at the `Device` component, that can decide if the device is individually rotated. 

And then pass that state value and the setter function to the `Device/Toolbar.jsx` component and make it functional. 
And inside the `Device` component, we can use both the redux store boolean value and this local state value and decide if the device needs rotation or now. 

Wdyt? 
I would prefer keeping it simple for easier development on top of this in the future. Please feel free to ping me on Discord if you have any question on this approach.

WayneRocha: @manojVivek that was a good ideia to simplify the code.
Wdyt about the changes?

manojVivek (approved):
This looks so much simple and easy to understand! 
Thank you so much for the contribution, looking forward to more like this!

manojVivek:
@all-contributors Please add @WayneRocha for code.

PR #936, opened 2023-05-14
Feature merge all screenshots

### LINKED ISSUE #917: Screenshot all preview devices

durgakiran:
# ✨ Pull Request
Add merge screenshot functionality to existing capture all screen screen function

### 📓 Referenced Issue

Fixes #917 

### ℹ️ About the PR

Adds merge all screens using buffer data merge. I am using `jpeg-js` library for proper decoding of data to jpeg images.
The merge on nodejs side should give us some performance benefits like not blocking main event loop. Also there is prefix input box which can be used to enter the prefix to be added for merged image while saving to fs. User has option to merge all the screens or not, below screenshot should give all clarity on options.
image

- Save each image separately option allows each webview to be captured and saved as jpeg image.
- Merge images option allows all webviews to be captured, merged and saved as jpeg image.
- When user chooses merge images options he must enter prefix (prefix to the merged jpeg image) otherwise not needed.
test 4-1684071131934

### DIFF
--- a/desktop-app/package.json
+++ b/desktop-app/package.json
@@ -117,6 +117,7 @@
     "electron-store": "^8.0.2",
     "electron-updater": "^5.2.3",
     "emitter": "^0.0.5",
+    "jpeg-js": "^0.4.4",
     "mousetrap": "^1.6.5",
     "os": "^0.1.2",
     "postcss": "^8.4.14",
--- a/desktop-app/src/common/image/bitmap.ts
+++ b/desktop-app/src/common/image/bitmap.ts
@@ -1,0 +1,261 @@
+/**
+ * Bitmap class to create image bitmaps
+ */
+import { WriteStream, createWriteStream } from 'fs';
+import jpeg from 'jpeg-js';
+
+interface RGBA {
+  r: number;
+  g: number;
+  b: number;
+  a: number;
+}
+
+interface FileOptions {
+  quality: number;
+  type: ImageType;
+}
+
+export type COLOR = RGBA;
+
+export interface BitmapData {
+  width?: number;
+  height?: number;
+  data?: ArrayBuffer;
+  color?: COLOR;
+}
+
+export enum FileType {
+  PNG = 'png',
+  JPEG = 'jpeg',
+  JPG = 'jpg',
+}
+
+export enum ImageType {
+  PNG = 1,
+  JPEG = 2,
+}
+
+export const TRANSPARENT_BLACK = { r: 0, g: 0, b: 0, a: 0 };
+
+export class Bitmap {
+  private width!: number;
+
+  private height!: number;
+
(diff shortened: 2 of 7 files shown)


durgakiran:
@manojVivek  one more step toward merge functionality. I have added `jpeg-js` dependency as a way to decode jpeg data which is need for image merging. Having this merge on server side gives us better performance ( I believe). Please let me know what you think.

manojVivek:
@durgakiran I just took look at this PR. Apologies for the delay. 
A couple of things:
- Is it possible to add the device name (and the dimensions?) on the top of the screenshots? It would make the merged image more useful for demonstrations.
- the UI hangs a bit while processing the images for large websites, probably because of the `jpeg-js` thing you mentioned above? I would suggest we try and make the merge on the main process instead of the renderer. If it's too hard, then we can keep that as an improvement separately. 


Overall, great work pulling this enormous feature out! 🚀 👏🏼

durgakiran:
> @durgakiran I just took look at this PR. Apologies for the delay. A couple of things:
> 
> * Is it possible to add the device name (and the dimensions?) on the top of the screenshots? It would make the merged image more useful for demonstrations.
> * the UI hangs a bit while processing the images for large websites, probably because of the `jpeg-js` thing you mentioned above? I would suggest we try and make the merge on the main process instead of the renderer. If it's too hard, then we can keep that as an improvement separately.
> 
> Overall, great work pulling this enormous feature out! 🚀 👏🏼

@manojVivek  The merging is entirely on main thread ( shouldn't block render thread) but let me investigate it further. It would be great if you provide any example website where it hangs.

manojVivek:
> The merging is entirely on main thread ( shouldn't block render thread)

Oh, my bad then. 

I tried `github.com` and the cursor goes to loading state for a couple of seconds and the following error is shown.


Screenshot 2023-05-30 at 4 55 03 PM

durgakiran:
@manojVivek  how many devices you are trying to capture at a time?

manojVivek:
The 3 default devices.

durgakiran:
> The 3 default devices.

@manojVivek 
I got the issue, it is happening because of jpeg-js configuration. On further investigation I found that Electron's Native Image has everything that we needed, so removed `jpeg-js` entirely.  On performance side I am trying to do some profiling and want to introduce worker threads if possible so that we don't have to block main thread at all.

On adding device name to the merged images, I still think we need to discuss it further. However this PR is pre-requiste to add the devices names. (Will work on that as different issue).

manojVivek:
> On adding device name to the merged images, I still think we need to discuss it further. However this PR is pre-requiste to add the devices names. (Will work on that as different issue).

Durga, I think we should have some kind of indication on which screenshot is for which device. Otherwise, the image would become less meaningful while sharing.

Issue #970, opened 2023-06-16
Switch to 100% zoom level when switching to the individual device layout


manojVivek:
When switching to the individual layout using the following icon:

<img width="496" alt="Screenshot 2023-06-16 at 11 48 28 AM" src="https://github.com/responsively-org/responsively-app/assets/1283424/6e4bbf8a-2235-4c89-820c-1cfc079051f3">

Change the zoom level to `100%` and restore it back to the original zoom level when switching back to the non-individual layout. 

Ideally, we should use two different zoom levels in the redux store:
1. non-individual layout zoom (current one, no change needed)
2. one for individual layouts (This should be implemented)

jibranabsarulislam:
on it! i just got it working on my end, I'll open a PR and tag this issue

PR #972, opened 2023-06-16
Set zoomFactor to 1 on individual, persisting column view zoomFactor

jibranabsarulislam:
# ✨ Pull Request

### 📓 Referenced Issue

#970 

### ℹ️ About the PR

add the `setZoom100` and `restoreZoom` reducers to set the `zoomFactor` to 1 and store `columnsZoomFactor`, then restores `columnsZoomFactor` respectively.

https://github.com/responsively-org/responsively-app/assets/70596906/709fed2f-06be-41da-8d56-48f036b0a208


desktop-app/src/renderer/components/Previewer/Device/index.tsx
```diff
@@ -199,8 +201,10 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
     if (!isIndividualLayout) {
       dispatch(setLayout(PREVIEW_LAYOUTS.INDIVIDUAL));
       setIndividualDevice(selectedDevice);
+      dispatch(setZoom100());
```
manojVivek, line 204: Instead of `restoreZoom` logic, can you try handling it in the `Device` component to use the `zoomFactor` when in non-individual mode and `individualZoomFactor` when in an individual mode? 

That way, the user set zoom levels for each of those modes will be retained. The user won't be annoyed about the 100% zoom if they don't like it. They can change it once and we'll use them the next time onwards.

manojVivek (changes requested):
Thanks for the quick implementation, shared my thoughts on the approach.

desktop-app/src/renderer/components/Previewer/Device/index.tsx
jibranabsarulislam: thanks again Manoj,

I've got it to persist the zoom factor used for the individual view as `individualZoomFactor`. should each device have their own zoomFactor store, or is the individual zoom applicable at a higher level to just the view?

manojVivek: One global `individualZoomFactor` should be enough to start with.

jibranabsarulislam: got it, thanks @manojVivek. I went with your second suggestion, resulting in global `individualZoomFactor` implemented and functioning

manojVivek (approved):
Works perfect!! Thank you for the contribution, looking forward to more like this. 💯

manojVivek:
@all-contributors Please add @jibranabsarulislam for code.

manojVivek:
@jibranabsarulislam Please join us on our Discord if you haven't already: https://responsively.app/join-discord

PR #1004, opened 2023-07-03
add some unit tests to DeviceManager function

Ehsanino82:
# ✨ Pull Request


themohammadsa (changes requested):
Hey @Ehsanino82. Thank you for raising the PR. 😄 

But unfortunately, the changes that you have made are to the legacy app of responsively, which is no longer being used. We have shifted to the newer version which is present in the directory `/desktop-app`. 

Fortunately, we do have a lot of [issues](https://github.com/responsively-org/responsively-app/issues) that need your help, you can pick up any of them. Hoping for more contributions from you! 😃

PR #1025, opened 2023-07-19
README: Improve the way to the downloads

amenk:
As a future user I want to find the downloads fast :-)
It took me a minute to find the correct location to download - this should reduce this to seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


manojVivek (approved):
Thank you so much for fixing these URLs. ❤️

manojVivek:
@all-contributors Please add @amenk for docs.

amenk:
Thanks for the quick processing and the awesome tool

Issue #1022, opened 2023-07-18
Last device view in Preview Layout Stacked mode on macOS being cut off


datverse:
Hi, 
When choice "Stacked mode" in Preview Layout. The Last device view being cut off, the footer being lost something.

My Environment:

- macOS: 13.4.1 
- ResponsivelyApp: 1.6.0

Please view my screenshot.

[https://postimg.cc/yDnp9VNk](https://postimg.cc/yDnp9VNk)

Thanks

EgeDogukan:
I have found the problem, currently working on it, will open a pull request soon.

manojVivek:
@EgeDogukan That's fantastic! Looking forward.

datverse:
Hi, please help us about this issue. We waiting for this.
❤️❤️❤️

datverse:
Hi @manojVivek ,
Does this issue will be fix in next release?
Thanks ❤️

ankitBytes:
Is this issue resolved yet?
If not I can fix it.

datgausaigon:
Hi @manojVivek ,
Can you look into help from @ankitBytes ?
Thanks 
🐻

happymvp:
@manojVivek why @EgeDogukan PR was ignored? Is this issue still valid?

PR #1033, opened 2023-08-02
fix: bottom screen cut off #1022

EgeDogukan:
# ✨ Pull Request

### 📓 Referenced Issue

#1022

### ℹ️ About the PR

Problem occurs in flex mode when a device which does not have ability to scroll is placed at the bottom. What I did is, I first check whether the device has a scrolling ability if not then I manipulate its height value. I first intended to make it scale with dpi value but noticed that not everything had one or you could change it manually. So, I settled with a value which probably will be okay for the most of the time. I can make a better solution with some help i.e. with scaling dpi (because there can be small white space left for some desktop screens since it does not scale with any constant, put in 2nd screen shot) but I feel like this is the best solution we can have right now. At least better than not seeing the bottom part of the page. 😅 

Note: This is my first contribution to any open source project, so feel free to point out where I can do better . 😄

Screenshot 2023-08-02 at 18 05 22

Screenshot 2023-08-02 at 18 14 41

### DIFF
--- a/desktop-app/src/renderer/components/Previewer/Device/index.tsx
+++ b/desktop-app/src/renderer/components/Previewer/Device/index.tsx
@@ -426,7 +426,9 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
     };
   }, [dispatch, isPrimary]);
 
-  const scaledHeight = height * zoomfactor;
+  const scaledHeight = device.isMobileCapable
+    ? height * zoomfactor
+    : (height + device.dpi * 17) * zoomfactor;
   const scaledWidth = width * zoomfactor;
 
   return (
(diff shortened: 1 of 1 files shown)


desktop-app/src/renderer/components/Previewer/Device/index.tsx
```diff
@@ -426,7 +426,9 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
     };
   }, [dispatch, isPrimary]);
 
-  const scaledHeight = height * zoomfactor;
+  const scaledHeight = device.isMobileCapable
+    ? height * zoomfactor
+    : (height + 30) * zoomfactor;
```
manojVivek, line 431: Any idea why it needs `30px` more only when `isMobileCapable`?

manojVivek (commented):
Apologies for the delay in reviewing this! 

Just a question on the solution for better understanding the problem.

EgeDogukan:
Problem occurs when the device at the bottom of the screen is not mobile capable. The ability to scroll on mobile device type, handles the problem and you can still perfectly see the bottom window. But on devices where you cannot scroll, bottom section of the window is unreachable when the said windows are on the bottom part of the app. So, we correct its height value there on devices where the problem occurs.

PR #1048, opened 2023-08-15
Fixed refresh docking issues

### LINKED ISSUE #999: Program window go width 100% when refreshing page
Hi,
when I have RA only on half screen and refresh page program window automatically go to full screen.

Operating system: Windows 11.

Eze-DP:
Fixes #999 refresh and docking issues. 

The problem is fixed by storing the window information when pressing the refresh button, and then applying it to the window when the electron.js ready-to-show function is called. 

https://github.com/responsively-org/responsively-app/assets/105628058/e3f37651-ddeb-477c-81fc-855ab8061e9d

### DIFF
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -173,11 +173,32 @@ const createWindow = async () => {
     )}`
   );
 
+  interface WindowBounds {
+    x: number;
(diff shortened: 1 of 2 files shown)


Eze-DP:
Any feedback please?

Issue #1050, opened 2023-08-16
App gets stuck when taking full page screenshot


gomorizsolt:
# 🐞 bug report

### ✍️ Description

Related issue: https://github.com/responsively-org/responsively-app/issues/1003

When taking full page screenshots of all/individual devices, the app gets stuck. Quick screenshot is not working either.

### 🕵🏼‍♂️ Is this a regression?

Could be, because the bug started to appear right after upgrading to the latest version (`v1.7.0`).

### 🔬 Minimal Reproduction

Add any device to the suite and press "Screenshot All WebViews" / "Full Page Screenshot" / "Quick Screenshot".

### 🌍 Your Environment

<pre><code>
Version: 1.7.0
Electron: 25.2.0
Chrome: 114.0.5735.134
Node.js: 18.15.0
V8: 11.4.183.23-electron.0
OS: Darwin arm64 22.5.0
</code></pre>

### 🔥 Exception or Error or Screenshot

![responsively_bug](https://github.com/responsively-org/responsively-app/assets/30864202/981d564b-18f9-4757-b3c6-41da4c7de1c9)

rei2232:
https://github.com/responsively-org/responsively-app/pull/1055

gomorizsolt:
@manojVivek I am still experiencing the same behaviour, after updating to the latest version.

PR #1055, opened 2023-08-29
Fix screenshot error from #1050

rei2232:
Fix issue https://github.com/responsively-org/responsively-app/issues/1050 

- When taking screenshot it gives:
```
Error occurred in handler for 'screenshot': TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:399:5)
    at validateString (node:internal/validators:163:11)
    at Object.join (node:path:1172:7)
    at quickScreenshot (/home/yung/Documents/responsively-app/desktop-app/src/main/screenshot/index.ts:63:25)
    at async EventEmitter.<anonymous> (node:electron/js2c/browser_init:2:88864) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```
- Which leads to Store(electron-store) schema of desktop-app/src/store/index.ts giving:
```
ResponsivelyApp encountered an error
Error: Config schema violation: `userPreferences` must be object
```
- And fixed this issue by spliting userPreference object from Store schema.

### DIFF
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -258,7 +258,7 @@ app.on('certificate-error', (event, _, url, __, ___, callback) => {
     return callback(true);
   }
   console.log('certificate-error event', url, BROWSER_SYNC_HOST);
-  return callback(store.get('userPreferences.allowInsecureSSLConnections'));
+  return callback(store.get('allowInsecureSSLConnections'));
 });
 
 app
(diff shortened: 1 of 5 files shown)


manojVivek:
@rei2232 But we mainly wanted to these preference items grouped behind the `userPreferences` hierarchy. I think the root cause is that there is no default value set for this object.

manojVivek:
I have taken care of this in #1059, so closing this PR.

Issue #1020, opened 2023-07-17
Create an alternative full-page screenshot algorithm that doesn't modify the device's viewport dimensions. 


manojVivek:
Ref: https://discord.com/channels/992277343408246825/992278054682497035/1130499133673844808

shred787:
Thanks for this Mano. It's clear that this is an important issue. I subbed to this thread.

PR #1058, opened 2023-09-03
Add option to choose not to modify the device's viewport dimensions

Mounir-Bouaiche:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes: #1020 

### ℹ️ About the PR

This is useful when you want to take a screenshot with the device dimensions being the minimum.

#### With:
iPhone 12 Pro-1693755429389

#### Without:
iPhone 12 Pro-1693755759659

### DIFF
--- a/desktop-app/src/main/screenshot/index.ts
+++ b/desktop-app/src/main/screenshot/index.ts
@@ -1,6 +1,6 @@
 /* eslint-disable promise/always-return */
 import { Device } from 'common/deviceList';
-import { ipcMain, shell, webContents } from 'electron';
+import { BrowserWindow, ipcMain, shell, webContents } from 'electron';
 import { writeFile, ensureDir } from 'fs-extra';
 import path from 'path';
 import store from '../../store';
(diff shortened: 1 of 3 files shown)


desktop-app/src/store/index.ts
```diff
@@ -148,6 +148,10 @@ const schema = {
         },
         default: {},
       },
+      onlyVisibleParts: {
+        type: 'boolean',
+        default: false,
```
manojVivek, line 153: Can we make the default as `true`? So that the current behaviour stays the same unless specifically modified in the settings.

desktop-app/src/main/screenshot/index.ts
```diff
@@ -43,7 +43,50 @@ const captureImage = async (
...
+      webPreferences: {
+        offscreen: true,
+        nodeIntegration: false,
+        contextIsolation: true,
+      },
+    });
+
+    try {
+      if (rect != null) {
+        console.log(rect);
```
manojVivek, line 77: Can you please remove this debug log line?

desktop-app/src/main/screenshot/index.ts
```diff
@@ -43,7 +43,50 @@ const captureImage = async (
...
+        screenShotDelegate.setContentSize(rect.width, rect.height);
+      }
+    } catch (ex) {
+      //
```
manojVivek, line 81: Maybe console.log the error here?

manojVivek (changes requested):
This is a very thoughtful feature and glad you implemented it!
Just a couple of minor thoughts, looks great otherwise!

PR #1071, opened 2023-09-10
Fix MacBook capitalization in deviceList.ts

jeffbowen:
Should be MacBook instead of Macbook. See https://www.apple.com/macbook-pro/


manojVivek (approved):
Good catch! thanks for the PR.

manojVivek:
@all-contributors Please add @jeffbowen for code.

PR #1081, opened 2023-09-18
Add ability to edit a custom device's name

### LINKED ISSUE #749: Support device renaming
# 🚀 Feature Request

### 📝 Description

It would probably a good idea to implement the possibility to rename a device. Everyone makes typo mistakes so it would be a good fast way to rename the device instead of delete it and make a new one.

It could something like a edit button in the device manager on the "active devices" site just like the delete button on the inactive site

### LINKED ISSUE #922: [FR] Ability to edit device name
# 🚀 Feature Request

### 📝 Description

After creating a custom device, you can't edit the device name.

You should be able to edit the device name.

lyncasterc:
# ✨ Pull Request

### 📓 Referenced Issue
- Resolves #922 
- Resolves #749

### ℹ️ About the PR
In `DeviceDetailsModal.tsx`:

- only disable name input if it is not a custom device.
- remove the note informing user that they can not change the custom device's name
- Disallow the ability to change a custom device's name to an existing device's name in the `handleAddDevice` function
  -  I added this because it is already the existing behavior when adding a new custom device, and because it creates issues when checking/unchecking custom devices with the same name because the logic in `DeviceLabel.tsx` is still using name instead of id.

### DIFF
--- a/desktop-app/src/renderer/components/DeviceManager/DeviceDetailsModal.tsx
+++ b/desktop-app/src/renderer/components/DeviceManager/DeviceDetailsModal.tsx
@@ -133,11 +133,8 @@ const DeviceDetailsModal = ({
               placeholder="My Mobile Device"
               value={name}
               onChange={(e) => setName(e.target.value)}
-              disabled={!isCustom || !isNew}
+              disabled={!isCustom}
             />
-            {!isNew && isCustom ? (
-              <p>Note: Device name cannot be modified once created!</p>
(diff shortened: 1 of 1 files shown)


manojVivek (approved):
This looks fantastic! 
Thank you so much for the contribution and looking forward to more!

manojVivek:
@all-contributors Please add @lyncasterc for code.

PR #1103, opened 2023-09-29
Update index.tsx

mohitthakur9901:
# ✨ Pull Request

### ℹ️ About the PR
navbar button color changed 

changed the navbar download button color because the same button on page has different color both button's functionality is same but color not

### DIFF
--- a/desktop-app/src/renderer/components/Button/index.tsx
+++ b/desktop-app/src/renderer/components/Button/index.tsx
@@ -48,8 +48,8 @@ const Button = ({
     hoverBg = 'hover:bg-slate-200';
     hoverBgDark = 'dark:hover:bg-slate-700';
   } else if (isPrimary) {
-    hoverBg = 'hover:bg-emerald-600';
-    hoverBgDark = 'dark:hover:bg-emerald-600';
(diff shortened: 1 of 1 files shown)


desktop-app/src/renderer/components/Button/index.tsx
```diff
@@ -48,8 +48,8 @@ const Button = ({
     hoverBg = 'hover:bg-slate-200';
     hoverBgDark = 'dark:hover:bg-slate-700';
   } else if (isPrimary) {
-    hoverBg = 'hover:bg-emerald-600';
-    hoverBgDark = 'dark:hover:bg-emerald-600';
+    hoverBg = 'hover:bg-slate-200';
```
manojVivek, line 51: I guess, we need to give them more emphasis if they're `primary` as compared to the colors above?

mohitthakur9901: I standardized the button color to slate for consistency since both buttons served the same function. This ensures a unified user experience.

manojVivek: The thing is the `Primary` variant is an emphasised version and so it is designed to have the brand colour bg.

PR #1140, opened 2023-10-22
[Feature] Added a Custom Titlebar

### LINKED ISSUE #189: [Feature] Add the File menu directly to the title bar
I really love this project and I have been using it for my project for quite some time now.

image

I think it would be a good idea to add the file menu directly into the status bar on Windows and Linux as it would help optimize the screen space available and it would amazing for devices with a smaller screen
Thanks in Advance.

ParamBirje:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes #189 

### ℹ️ About the PR

As suggested by @Kidcredo , The [custom-electron-toolbar](https://www.npmjs.com/package/custom-electron-titlebar) package was added to the project to tweak the default toolbar. This made it possible to make the File menu and the Title be in the same space which optimizes the screen space available and more accessible for devices with smaller screen. 

The app's brand color is now used by the custom titlebar.

Previous titlebar

previous_titlebar

Custom titlebar

new_titlebar

### DIFF
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -104,11 +104,11 @@ const installExtensions = async () => {
     .catch(console.log);
 };
 
-// Custom titlebar config
+// Custom titlebar config for windows
 const customTitlebarStatus = store.get(
   'userPreferences.customTitlebar'
 ) as boolean;
-if (customTitlebarStatus) {
+if (customTitlebarStatus && process.platform === 'win32') {
   setupTitlebar();
 }
 
@@ -131,9 +131,11 @@ const createWindow = async () => {
     width,
     height,
     icon: getAssetPath('icon.png'),
-    titleBarStyle: customTitlebarStatus ? 'hidden' : 'default',
+    titleBarStyle:
+      customTitlebarStatus && process.platform === 'win32'
+        ? 'hidden'
+        : 'default',
     webPreferences: {
-      sandbox: !customTitlebarStatus,
       preload: app.isPackaged
         ? path.join(__dirname, 'preload.js')
         : path.join(__dirname, '../../.erb/dll/preload.js'),
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -12,6 +12,10 @@ import path from 'path';
 import { app, BrowserWindow, shell, screen, ipcMain } from 'electron';
 import { autoUpdater } from 'electron-updater';
 import log from 'electron-log';
+import {
+  setupTitlebar,
+  attachTitlebarToWindow,
+} from 'custom-electron-titlebar/main';
 import cli from './cli';
 import { PROTOCOL } from '../common/constants';
 import MenuBuilder from './menu';
@@ -100,6 +104,8 @@ const installExtensions = async () => {
     .catch(console.log);
 };
 
+setupTitlebar();
+
 const createWindow = async () => {
   windowShownOnOpen = false;
   await installExtensions();
@@ -119,7 +125,9 @@ const createWindow = async () => {
     width,
     height,
     icon: getAssetPath('icon.png'),
+    titleBarStyle: 'hidden',
     webPreferences: {
+      sandbox: false,
       preload: app.isPackaged
         ? path.join(__dirname, 'preload.js')
         : path.join(__dirname, '../../.erb/dll/preload.js'),
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -104,7 +104,13 @@ const installExtensions = async () => {
     .catch(console.log);
 };
 
-setupTitlebar();
+// Custom titlebar config
+const customTitlebarStatus = store.get(
+  'userPreferences.customTitlebar'
+) as boolean;
+if (customTitlebarStatus) {
+  setupTitlebar();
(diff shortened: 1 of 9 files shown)


manojVivek:
@ParamBirje Sorry for the delayed response, can you please try some subtle color for the background instead of the brand green? 
Something default matching the windows theme should be perfect.

ParamBirje:
Hey @manojVivek , I have made the required changes to the titlebar making it's background color subtle and feel default. Here's how it looks.

image

manojVivek:
Looks goods, also, is it possible to make this a user preference? 

So that it doesn't affect users who like it the current way.

ParamBirje:
The custom titlebar actually builds itself during the creation of BrowserWindow, I don't think we can change it in realtime. We could probably save the user preference and on the next restart, load the preferred titlebar. I'll open a new issue/feature for changing the titlebar preferences, if that sounds good?

manojVivek:
Yes, applying the change on restart sounds good (we can add a not below the user preference option indicating that). 

Yes, a new issue is fine, but I would like to merge both of these together, as we wouldn't want to surprise the users with something they didn't expect.

ParamBirje:
Could you please give me some pointers to how the storage part be implemented? We can't use `window.electron.store` because the window doesn't exist yet. Should it be done using the file system (fs) API in electron or something else?

manojVivek:
@ParamBirje You don't need any extra package for this. 

You can directly access the user preferences in the main process like this: https://github.com/responsively-org/responsively-app/blob/34c3b903f7e843413e403cdb35be96697a96d945/desktop-app/src/main/main.ts#L261

ParamBirje:
Hey @manojVivek ,

Thank you for the assist on electron store, saved a lot of time. Coincidently the package that I had found was `electron-store` itself, and then noticed it was already implemented in the project. The Titlebar Switch feature which switches between the new custom titlebar and the default titlebar has been implemented.

Custom Titlebar (default preference)

custom_titlebar

Default Titlebar

default_titlebar

Let me know if there are any changes required.

desktop-app/src/main/main.ts
```diff
@@ -119,7 +125,9 @@ const createWindow = async () => {
     width,
     height,
     icon: getAssetPath('icon.png'),
+    titleBarStyle: 'hidden',
     webPreferences: {
+      sandbox: false,
```
manojVivek, line 130: Why do we want to disable the `sandbox` option?

desktop-app/src/main/main.ts
```diff
@@ -119,7 +125,9 @@ const createWindow = async () => {
     width,
     height,
     icon: getAssetPath('icon.png'),
+    titleBarStyle: 'hidden',
```
manojVivek, line 128: Also, we want to ensure this doesn't affect Mac and Linux versions of the app, do you have any ideas how it works there after this change?

desktop-app/src/main/main.ts
ParamBirje: The `custom-electron-titlebar` mentioned it in their [documentation](https://github.com/AlexTorresDev/custom-electron-titlebar/wiki/Get-Started), for the library to function properly, `sandbox` should be `false`.

However, I tested the functionality by **omitting** the `sandbox` line, and found out, the custom titlebar works just fine and encountered no issues yet. 

Should I keep this line anyways?

desktop-app/src/main/main.ts
ParamBirje: I don't think this would affect Mac/Linux versions, as that option disables default taskbar styles and results in a full-size content window.

ParamBirje:
Hey @manojVivek ,

Here's the toggle implemented in the Settings panel.

toggle_in_settings

desktop-app/src/main/main.ts
manojVivek: If it works without this, then I would suggest we remove this, as this is option is a compromise on security.

desktop-app/src/main/main.ts
manojVivek: I just tried this code on Mac and it affects the title bar color.

Preference disabled(default):
Screenshot 2023-10-29 at 11 39 21 AM

Preference enabled:
Screenshot 2023-10-29 at 11 38 42 AM


Is it possible to selectively apply this by checking, only when the platform is windows?

manojVivek: Try moving the styles to a new css file and optionally import based on UA data?
Screenshot 2023-10-29 at 11 53 53 AM

ParamBirje:
Hey @manojVivek ,

I have made changes to the custom titlebar's implementation such that only if the platform is Windows, the feature can be used. The UI which toggles the custom titlebar in the Settings, is now only visible to Windows users. I have also moved the custom titlebar CSS to a new file as directed.

Here's how I am checking platforms in `main.ts` and `preload.ts`

preload_main

And in the React files (process is not exposed here)

tsx_files

[The `sandbox: false` is now removed.]

manojVivek (approved):
This is looking good! Thanks for patiently addressing all the comments! ♥️

ParamBirje:
It's been great dealing with issues and solutions back and forth. Thank you for letting me have a chance at contributing to a great project, with you and fellow maintainers doing a great job at keeping it that way! 😄

Issue #1149, opened 2023-11-03
Rulers & Guides


astuanax:
# 🚀 Feature Request

### 📝 Description

Should we have rulers and guides as an option?
I guess designers would love that. 

### ✨ Describe the solution you'd like

I already made a poc for rulers and guides, but I think we can definitively extend it with 
columns and grids.

Here is a peek at the feature:

https://github.com/responsively-org/responsively-app/assets/1730624/80ffa8e8-0b8b-46fc-a8eb-595942fc5acc

astuanax:
The rulers and guides features is starting to look usable. 

If anyone is interested, let me know whether you prefer the following:

- Rulers enable on all screens or one (right now only 1 is possible)
- Guides need to be saved for the next time?

Here is another peek:

https://github.com/responsively-org/responsively-app/assets/1730624/86e0a7e2-b625-46a6-8c5e-f47e410e4b31

PR #1162, opened 2023-11-14
Guides & Rulers

astuanax:
# ✨ Pull Request

### 📓 Referenced Issue 

#1149 Adds persistent guides and rulers

### ℹ️ About the PR

- Adds vertical and horizontal rulers
- Adds persistent guides which can be dragged, vertical and horizontal
- Adds keyboard short alt-r to show/hide rulers for all webviews

- Test with different screens resolutions
- Test with different views: single/many
- Close and reopen application after creating guides: they should remain visible
- Switch between night mode /regular mode


desktop-app/src/store/migrations.ts
```diff
@@ -88,4 +89,7 @@ export const migrations = {
       },
     ]);
   },
+  '1.2.2': (store: Store) => {
+    // Do we need a new migration???
```
manojVivek, line 93: I don't think we need any migrations as we don't change any existing schema fields.

desktop-app/src/store/index.ts
```diff
@@ -138,6 +138,27 @@ const schema = {
...
+            is_vertical: {
+              type: 'boolean',
+            },
+            resolution: {
+              type: 'string',
+            },
+          },
+          default: {},
+        },
+      },
```
manojVivek, line 161: Can you add a default for the `userPreferences.guides` field as a `[]` too? 

As on the first run, I'm getting this error:
Screenshot 2023-11-25 at 7 34 02 PM

desktop-app/src/renderer/components/Previewer/Guides/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+                    }),
+                  {
+                    resolution: `${width}x${height}`,
+                    is_vertical: false,
+                    positions: guides.filter((x) => x > 0),
+                  },
+                ]);
+              }}
+            />
+            <Guides
```
manojVivek, line 148: Any idea why the numbers are getting cut off in iPhone 12 Pro:
Screenshot 2023-11-25 at 7 40 39 PM

Looks good on other wider screen devices though.

desktop-app/src/store/index.ts
```diff
@@ -198,7 +219,7 @@ const schema = {
   },
   homepage: {
     type: 'string',
-    default: 'https://www.google.com/',
+    default: 'https://www.example.org/',
```
manojVivek, line 222: Is this some debug change got into the commit?

desktop-app/src/renderer/components/Previewer/Device/index.tsx
```diff
@@ -450,33 +515,58 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
...
+          enabled={rulerEnabled(`${width}x${height}`)}
+          defaultGuides={window.electron.store
+            .get('userPreferences.guides')
+            .flatMap((x: any) => x)
+            .filter((x: DefaultGuide) => {
+              return x.resolution === `${width}x${height}`;
+            })}
         />
+        <div style={{}} className="bg-slate-200 dark:bg-slate-800">
+          <webview
```
manojVivek, line 547: It looks like the default `white` background style is not working any more? 

Screenshot 2023-11-25 at 7 42 46 PM

desktop-app/src/renderer/components/Previewer/Device/Toolbar.tsx
```diff
@@ -124,14 +126,24 @@ const Toolbar = ({
...
     setRotated(!rotated);
     onRotate(!rotated);
   };
 
   return (
-    <div className="flex items-center justify-between gap-1">
-      <div className="my-1 inline-flex max-w-[75%] items-center gap-1 overflow-x-auto">
+    <div
+      className="flex items-center justify-between gap-1"
+      style={{ paddingLeft: '30px' }}
```
manojVivek, line 144: Can we not add the`paddingLeft` to the device name row and toolbar row? 
We can only add it for the device row as this adds an unnecessary indentation for the title and the toolbar row. 

Also, can we add the margin to the device only when the ruler is activated? As it doesn't make a lot of sense to have this empty space all other times as well.

manojVivek (changes requested):
This is such an essential utility and great work implementing it. 🚀 

Did review the code and also a round of testing and added the observations as comments on the code.

desktop-app/src/renderer/components/Previewer/Device/Toolbar.tsx
astuanax: I tried that but that means the screens are jumping 30px all the time.

Changing position of the view messes with muscle memory.

The only thing we could try is reducing the 30 px to less, but as I also tried that, it makes it less readable. 

I don't think the space is an issue. Only at first. After a day, you wont notice it anymore. 

From a ux point, moving elements all the time is not a good option, layout instability is an annoying issue. If you have a better idea, I am all ears.

desktop-app/src/store/migrations.ts
astuanax: Yes, leftover from early ideas. 
Removed.

desktop-app/src/store/index.ts
astuanax: Yes, updated.

desktop-app/src/store/index.ts
astuanax: ok, done.

manojVivek (approved):
Looking great now!

sam98brown:
Is this feature usable?

manojVivek:
It is not live yet, waiting on a related bug to be addressed. It should be out in a week or so.

Issue #1234, opened 2024-05-15
Add "ScrollToTop" Button for User Enhancement


codercake:
# 🚀 Feature Request

### 📝 Description

Currently, the application lacks a convenient way for users to quickly navigate back to the top of the page, especially when browsing through lengthy content. Implementing a "Scroll to Top" button would greatly enhance user experience by providing a simple and efficient means to return to the top of the page with just one click.

### ✨ Describe the solution you'd like

- Add a small, unobtrusive button at the bottom-right corner of the screen that becomes visible as the user scrolls down the page.
- When clicked, the button smoothly scrolls the page back to the top.
- Ensure the button remains responsive and accessible across different screen sizes and devices.

### ✍️ Describe alternatives you've considered

One alternative could be implementing a keyboard shortcut for scrolling to the top, but this may not be as intuitive or accessible to all users compared to a visible button. Some browsers already have a built-in "scroll to top" functionality when users click on the browser's scrollbar, but having a dedicated button within the application would provide a consistent experience across different platforms.

Shashan17J:
Hi, I would like to work on this issue. Could you please assign it to me?

manojVivek:
@Shashan17J Please proceed with the implementation, we don't assign issues.

PR #1244, opened 2024-05-30
Added Scroll to top feature

sagarhedaoo:
# ✨ Pull Request

### 📓 Referenced Issue

Link: https://github.com/responsively-org/responsively-app/issues/1234

Fixes: #1234 

### ℹ️ About the PR

I implemented the Scroll to top feature in the webview as well as in the toolbar. The Button disappears when the user is at the top of the page and when user goes down then it appears.
Upon clicking it scrolls back to the top.
Same for the toolbar button that I added/

I used command yarn test to test the code before pushing it with all the test cases passed.

Screen Recording 2024-05-30 at 2 56 04 AM


desktop-app/src/renderer/components/Previewer/Device/index.tsx
```diff
@@ -495,6 +495,61 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
   const scaledHeight = height * zoomfactor;
   const scaledWidth = width * zoomfactor;
 
+  useEffect(() => {
+    const webview = ref.current;
```
manojVivek, line 499: I would say, we don't need this button inside the view. We don't want to pollute the user space with the app code. 

The icon in the toolbar is good enough and this just feels like a redundant cta. Can you please remove this?

manojVivek (changes requested):
@sagarhedaoo Thank you so much for this contribution, just a minor change request. Looks great otherwise!

desktop-app/src/renderer/components/Previewer/Device/index.tsx
manojVivek: Took a stab at it myself.

manojVivek:
@sagarhedaoo Thanks again for your effort on this! ❤️

manojVivek:
@all-contributors Please add @sagarhedaoo  for code.

Issue #1258, opened 2024-07-01
Export/import option missing again in 1.12.0


kkmuffme:
# 🐞 bug report

### ✍️ Description

Export/import doesn't exist anymore in 1.12.0
Was implemented in https://github.com/responsively-org/responsively-app/pull/657

### 🕵🏼‍♂️ Is this a regression?

Yes

### 🌍 Your Environment

1.12.0

artnerdnet:
is this a good first issue? can I take this?

manojVivek:
Yes, please go ahead and let me know if you have any questions.

violetadev:
I see that all this is in the legacy code, so we must migrate it to the new app. I will add it in the settings, on top of the SAVE button:

![image](https://github.com/user-attachments/assets/48dd9c87-6e38-48c0-8fa4-652e2ecf1da7)

manojVivek:
Yes, that sounds perfect!

PR #1268, opened 2024-07-28
add import and export capabilities for devices and suites

violetadev:
# ✨ Pull Request

### 📓 Referenced Issue
Fixes: #1258

### ℹ️ About the PR

This PR reintroduces import/export capabilities in order to users to back up their settings and import them back. 

https://github.com/user-attachments/assets/f42be76d-f7db-4483-857a-74b2aeb5c966


desktop-app/src/renderer/store/features/device-manager/index.ts
```diff
@@ -70,6 +70,12 @@ export const deviceManagerSlice = createSlice({
       state.activeSuite = action.payload.id;
       window.electron.store.set('deviceManager.previewSuites', suites);
     },
+    addSuites(state, action: PayloadAction<PreviewSuite[]>) {
+      window.electron.store.set('deviceManager.previewSuites', []);
+      state.suites = action.payload;
+      state.activeSuite = action.payload[0].id;
+      window.electron.store.set('deviceManager.previewSuites', action.payload);
```
manojVivek, line 77: This looks like it will override the existing suites as well, wouldn't that result in a loss of data if the user already had any custom suites?

.all-contributorsrc
```diff
@@ -739,6 +739,15 @@
       "contributions": [
         "code"
       ]
+    },
```
manojVivek, line 742: You can remove this, as it will be autopopulated by allContributors bot that updates the README as well.

desktop-app/src/renderer/components/FileUploader/FileUploader.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { FileUploader, FileUploaderProps } from './FileUploader';
+import { useFileUpload } from './hooks';
+
+jest.mock('./hooks');
+
+const mockHandleFileUpload = jest.fn();
+const mockHandleUpload = jest.fn();
+const mockResetUploadedFile = jest.fn();
+
+describe('FileUploader', () => {
```
manojVivek, line 12: Really appreciate the unit tests! 🫡

manojVivek (changes requested):
You pulled it off so nicely!! 👏🏼 
Just a couple of minor things, looks great otherwise!

manojVivek:
Also on the UI layout, shall we try the following placement for the icons: 
Screenshot 2024-07-29 at 9 22 29 AM

And remove the dividers so that it will indicate that both suites and devices will be exported/imported.

desktop-app/src/renderer/components/DeviceManager/PreviewSuites/ManageSuitesTool/ManageSuitesTool.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return (
+    <>
+      <div className="flex flex-row content-end justify-end">
+        <Button
+          data-testid="upload-btn"
+          className="aspect-square w-12 rounded-full hover:!bg-slate-500"
+          onClick={() => setOpen(true)}
+        >
+          <Icon
+            icon="mdi:folder-upload"
```
manojVivek, line 51: I'm leaning towards `uil:import` and `uil:export` for the icons? 
As the current ones look more like a folder import/export to me. Wdyt?

violetadev:
> Also on the UI layout, shall we try the following placement for the icons: Screenshot 2024-07-29 at 9 22 29 AM
> 
> And remove the dividers so that it will indicate that both suites and devices will be exported/imported.

yes, this was my first approach, but there are a few issues with this:
1. The import / export refers to importing not only suites, but also custom devices created by the user, and that title refers to the suites. 
2. If I add it there, then I have to contaminate the PreviewSuites component with the setCustomDevices (which is unrelated) and then do prop drilling to the new component because we don't have a disptach action for it, so I need to pass it as prop. 

I suggest instead we do something like this to improve the user experience (I did this over the screenshot and it's not very polished but so you get the idea):
Screenshot 2024-07-29 at 09 52 35


1. Adjust font sizes and add accordion for each of the sections: suites, default devices and custom devices
2. Move the title from the top to under the x space, so we can align any settings CTAs we have that are global to this screen

if you are ok with this I can implement it in this same PR or open a new one following up (which one do you prefer?)

desktop-app/src/renderer/components/FileUploader/FileUploader.test.tsx
violetadev: I wanted to add more on the ManageSuitesTool but I am having issues with the electron store, I can't seem to mock it. Do you have any idea how to mock it properly?

This is the error:
```
FAIL  src/renderer/components/DeviceManager/PreviewSuites/ManageSuitesTool/ManageSuitesTool.test.tsx
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'store')

      4 | import { sanitizeSuites } from './utils';
      5 |
    > 6 | const activeDeviceIds: string[] = window.electron.store.get(
        |                                                   ^
      7 |   'deviceManager.activeDevices'
      8 | );
      9 |

```

desktop-app/src/renderer/store/features/device-manager/index.ts
violetadev: sure, should we add also a button to reset the whole thing so they can go to the initial clean state as well?

violetadev:
Now instead of replacing, it merges whatever was there with the new. 

https://github.com/user-attachments/assets/e58ce599-05fb-4933-afb3-7ba966fea0ee



Also implements accordion and rearranges UI

https://github.com/user-attachments/assets/fe99c677-bbd1-4d72-a29d-bfbb0ea18d8c

manojVivek:
> > Also on the UI layout, shall we try the following placement for the icons: Screenshot 2024-07-29 at 9 22 29 AM
> > And remove the dividers so that it will indicate that both suites and devices will be exported/imported.
> 
> yes, this was my first approach, but there are a few issues with this:
> 
> 1. The import / export refers to importing not only suites, but also custom devices created by the user, and that title refers to the suites.
> 2. If I add it there, then I have to contaminate the PreviewSuites component with the setCustomDevices (which is unrelated) and then do prop drilling to the new component because we don't have a disptach action for it, so I need to pass it as prop.
> 
> I suggest instead we do something like this to improve the user experience (I did this over the screenshot and it's not very polished but so you get the idea): Screenshot 2024-07-29 at 09 52 35
> 
> 1. Adjust font sizes and add accordion for each of the sections: suites, default devices and custom devices
> 2. Move the title from the top to under the x space, so we can align any settings CTAs we have that are global to this screen
> 
> if you are ok with this I can implement it in this same PR or open a new one following up (which one do you prefer?)

Yes, I like the described approach. Let's go with that.

desktop-app/src/renderer/components/FileUploader/FileUploader.test.tsx
manojVivek: Maybe we need to add a mock `electron` object to the window before the test runs? 

Something like this:

electron-mock.ts
```
const store = {};
window.electron = {
  store: {
    get: (key: string) => {
        return store[key];
    },
    set: (key: string, val: any) => {
       store[key] = val;
    }
  }
};
```

And then import this file at the top of the test file?

(haven't thought this through, just a quick suggestion)

manojVivek:
Merging this in now. You pulled it off so great for your first commit to the repo! @violetadev 👏🏼 

Thank you so much! Looking forward to more like this in the future.

manojVivek:
@all-contributors Please add @violetadev for code.

PR #1277, opened 2024-08-07
DPI to DPR

### LINKED ISSUE #1257: "Device DPI" is actually Device Pixel Ratio (DPR) and should allow decimals
# 🐞 bug report

### ✍️ Description

"Device Manager" - the "Device DPI" isn't actually "DPI" but "DPR" (Device Pixel Ratio)
How I know? The values used in the presets (1 and 2) would mean a screen is unusable since the resolution is so low. Common values for DPI are much higher, in the 70+ range.
Please rename this.

Additionally, it should allow decimals, since a DPR of 1.5 is rather common too and often behaves differently than 2.

### 🕵🏼‍♂️ Is this a regression?

No

### 🌍 Your Environment

1.12.0

wp043:
# ✨ Pull Request

Fixes #1257 

### ℹ️ About the PR

According to the bug report, changed the field DPI in the custom device list to DPR. Also allows floating number by parseFloat instead of parseInt and I've changed the step of the DPR to 0.1 instead of the default 1 to allow users adjust the DPR more easily.

dpr-decimal


desktop-app/src/store/index.ts
```diff
@@ -104,7 +104,7 @@ const schema = {
             type: {
               type: 'string',
             },
-            dpi: {
```
manojVivek, line 107: We need to add a migration for this rename as well, to handle the config files on existing installations. 

Migrations exist in this file: https://github.com/responsively-org/responsively-app/blob/main/desktop-app/src/store/migrations.ts

wp043:
```typescript
'1.2.2': (store: Store) => {
    // Migrate dpi to dpr in custom devices
    try {
      const previousCustomDevices: any[] = store.get(
        'deviceManager.customDevices'
      ) as any[];
      const newCustomDevices: Device[] = previousCustomDevices.map((device) => {
        const newDevice = {
          ...device,
          dpr: device.dpi !== undefined ? device.dpi : device.dpr,
        };
        delete newDevice.dpi;
        return newDevice as Device;
      });
      store.set('deviceManager.customDevices', newCustomDevices);
      console.log(
        'Migration for 1.2.2 successful',
        store.get('deviceManager.customDevices')
      );
    } catch (e) {
      console.log('Migration for 1.2.2 failed', e);
    }
  },
```

Is this a valid approach for migration following the flow of existing code?

manojVivek: Yes, looking good. Just make sure the version number is `1.13.0` as that will be the next release version.

violetadev (approved):
🚀

manojVivek:
@all-contributors Please add @wp043 for code.

Adding again as the other PR was in merge conflict.

Issue #1279, opened 2024-08-07
Unable to edit information of custom device after duplicate name error


wp043:
# 🐞 bug report

### ✍️ Description
In device manager, when we hit the error of duplicate device name, after the pop up, none of the inputs can be modified, even if exiting the window and add again.

### 🕵🏼‍♂️ Is this a regression?
No

### 🔬 Minimal Reproduction
Create a custom device in device manager first. Then add another one that has the same name as the previous one. When click on "Add", the error pops up. After that, try selecting the fields and edit the names.

### 🌍 Your Environment
<pre><code>
App Version: 1.13.0
Electron Version: 27.0.4
Chrome Version: 118.0.5993.129
Node.js Version: 18.17.1
V8 Version: 11.8.172.17-electron.0
OS: Windows_NT x64 10.0.22631
</code></pre>

### 🔥 Exception or Error or Screenshot
![error](https://github.com/user-attachments/assets/2f2819cc-6c84-49df-b9ee-9c58cbe37a8a)

pranavithape:
what you saying i can edit those values bro

pranavithape:
Wait May Be You Are Right But Partially , Values editable with arrows but selection frozen

pranavithape:
Yeh I confirmed there is error you cant select value and edit it freezes

PR #1282, opened 2024-08-09
Unable to edit information ...  #1279: fixed focus issue - it electro…

pranavithape:
There issue opened at : Unable to edit information of custom device after duplicate name error #1279

Error Is Located In Electron Specifically Its Focus Function , It Electron Feature But It Making User Experience Bad Is Input field's  Error So We Fixed By Adding Little Code In Main


desktop-app/src/main/main.ts
```diff
@@ -177,6 +177,30 @@ const createWindow = async () => {
     )}`
   );
 
+  const isWindows = process.platform === 'win32';
```
violetadev, line 180: LGTM but @manojVivek can you take a look too? I am not able to test it on windows, I tried it on mac and it doesn't break anything.

violetadev:
@all-contributors Please add @pranavithape for code.

Issue #837, opened 2023-02-27
App calls focus each page refresh 


mvaneijgen:
# 🐞 bug report

### ✍️ Description

While developing websites I got several windows open. I've BrowserSync running which automatically refreshes the page when something is saved. Since version 1 Responsively calls focus the the app when any pages is refreshed. I'll be updating my todolist and slowing all pages get refreshed one by one and each time the app jumps to the front while I wanted to work in let's say Safari. 

### 🕵🏼‍♂️ Is this a regression?

Don't 'call focus' (bring Responsively to the front) when a page has refreshed. 

### 🔬 Minimal Reproduction

Open any other app. Put the following meta tag into  <head> of the website loaded  in Responsively to have the browser refresh every 5 seconds. Bring the other app to the front (in focus) and notice Responsivly 'calling focus' (jumping in front) every 5 seconds. 

### 🌍 Your Environment

<pre><code>
Responsively Version 1.0.3 (1.0.3)

Chip Apple M2 Pro
Memory 16 GB
macOS Ventura 13.2.1
</code></pre>

manojVivek:
@mvaneijgen Do you run the production build or development mode of Responsively? 

This behaviour shouldn't happen if you're using the production build downloaded from: https://responsively.app/download/

Can you please confirm how you're running the app?

mvaneijgen:
I don't know which version I had downloaded (probably from Github), but I've downloaded the 'production' version now from your link and the same is happening. Here a screen recording with Responsively v0.9 vs v1 with the <meta http-equiv="refresh" content="5"> notice it coming to front every ~5 seconds 

[![https://youtu.be/3H88iywI-6Q](https://img.youtube.com/vi/3H88iywI-6Q/0.jpg)](https://www.youtube.com/watch?v=3H88iywI-6Q)

manojVivek:
That's strange. 
Do you embed browser sync in the web page that you're working on?
In any case, would it be possible for you to share a webpage that reproduces this?

mvaneijgen:
To exclude BrowserSync from causing the issue, I've just opened a index.html file with only the following meta tag in it     <meta http-equiv="refresh" content="5"> and opened that. Which also causes the issue. 

1. create index.html file 
2. Add <meta http-equiv="refresh" content="5">
3. Open the file in VS code and launch Live Server to open a server at the location of file https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer
4. Open any other app then Responsively

[...]

PR #1295, opened 2024-09-08
Fixes #1184

### LINKED ISSUE #1184: Hot reload forces app to become front window
# 🐞 bug report

### ✍️ Description

Given that I have a hot reloadable page loaded in the Responsively App, and the app is not currently the front window, when the page hot reloads, the Responsively app automatically becomes the front window, disrupting my work.

### 🕵🏼‍♂️ Is this a regression?

Unsure

### 🔬 Minimal Reproduction

- Load a hot reloadable page in Responsively
- Open your code editing program to the site that's currently loaded
- Make an edit in your code and save
- The Responsively app will automatically become the front window 

### 🌍 Your Environment

<pre><code>
Version: 1.10.0
Electron: 27.0.4
Chrome: 118.0.5993.129
Node.js: 18.17.1
V8: 11.8.172.17-electron.0
OS: Darwin arm64 23.1.0
</code></pre>

violetadev:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes: #1184
Fixes: #837 

### ℹ️ About the PR

This PR adds a variable to pick up when the app has been initiated, so it only comes to the front when it is being opened for the first time, but not on refresh.

STR: 
- Open the app
- Open a page with hot loading (in this case, it's a simple html page with live server (VS code extension)
- Make a change, the responsively app should not be focused 

https://github.com/user-attachments/assets/c8234d3d-6736-4935-8e5a-37412d365936

### DIFF
--- a/desktop-app/src/main/main.ts
+++ b/desktop-app/src/main/main.ts
@@ -11,6 +11,7 @@
 import path from 'path';
 import { app, BrowserWindow, shell, screen, ipcMain } from 'electron';
 import { setupTitlebar } from 'custom-electron-titlebar/main';
+import { useState } from 'react';
 import cli from './cli';
 import { PROTOCOL } from '../common/constants';
 import MenuBuilder from './menu';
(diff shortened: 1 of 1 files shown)


manojVivek (approved):
Lgtm! 

So glad that you could reproduce and make a fix, I was struggling to reproduce this consistently when I tried. 🎉

Issue #1298, opened 2024-09-09
Performance Improvement


aryan262:
# 🚀 Feature Request

### 📝 Description

Currently, the useEffect inside the URLOpenerNonChrome and URLOpenerChrome components contains asynchronous logic that is executed on every render. This can lead to unnecessary re-renders, especially if the component is rendered multiple times. By optimizing the code with useCallback and useMemo, we can improve the performance of the application, ensuring that certain computations are only executed when necessary.

### ✨ Describe the solution you'd like

1. Refactor the asynchronous logic into a useCallback:

- The logic for querying tabs and checking the custom protocol URI should be wrapped in useCallback to prevent unnecessary recalculations on each render.

2. Utilize useMemo to optimize dependent computations:

- Any computations inside useEffect that depend on props or state can be memoized with useMemo to improve performance and reduce redundant operations.

Fixes: https://github.com/responsively-org/responsively-app/issues/1298

PR #1300, opened 2024-09-09
Performance Improvement

aryan262:
🚀 Feature Request
📝 Description
Currently, the useEffect inside the URLOpenerNonChrome and URLOpenerChrome components contains asynchronous logic that is executed on every render. This can lead to unnecessary re-renders, especially if the component is rendered multiple times. By optimizing the code with useCallback and useMemo, we can improve the performance of the application, ensuring that certain computations are only executed when necessary.

✨ Describe the solution you'd like
Refactor the asynchronous logic into a useCallback:
The logic for querying tabs and checking the custom protocol URI should be wrapped in useCallback to prevent unnecessary recalculations on each render.
Utilize useMemo to optimize dependent computations:
Any computations inside useEffect that depend on props or state can be memoized with useMemo to improve performance and reduce redundant operations.

Fixes: https://github.com/responsively-org/responsively-app/issues/1298

### DIFF
--- a/browser-extension/src/popup.js
+++ b/browser-extension/src/popup.js
@@ -27,27 +27,22 @@ const isChrome = () => {
 
 const URLOpenerNonChrome = () => {
   const [status, setStatus] = useState("loading");
-  useEffect(() => {
-    (async () => {
-      const [tab] = await window.browser.tabs.query({
-        currentWindow: true,
-        active: true,
-      });
-      if (!tab || !tab.url) {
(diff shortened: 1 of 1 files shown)


violetadev:
Thanks for contributing, I will take a look.

Please add the following to the PR description, it will link to the issue you opened:

Fixes:  #1298

violetadev:
looks great, thank you for the refactor!!

PR #1313, opened 2024-09-21
Add a notification component

violetadev:
# ✨ Pull Request

### 📓 Referenced Issue

This is a notification component that is being added under the "..." menu on the toolbar.
The idea is  that we have a space to add important information, in this case I have added a version check, to let the user know when there is a new version available (I noticed a lot of users asking about the new features or bug fixes). This checks every 2 weeks for a new version on github.

There are two parts of this PR; the notification system, with its own hook and UI and the version check with its own hook.

# Tech Design
responsively-app drawio

# Notification UI
https://github.com/user-attachments/assets/31142440-be13-429c-8d45-e7c629794dec

# Empty State
Screenshot 2024-09-21 at 14 24 17

To test this, you will need to have an older version of the app, in order to see the notification for the new app download message.


manojVivek:
@violetadev Apologies for the delay, this looks good overall, but the only concern is that it would be great if we could leverage the native app updater into the workflow instead of taking the users to the website.

Here's the reference code to the app updater: https://github.com/responsively-org/responsively-app/blob/main/desktop-app/src/main/main.ts#L242C9-L242C19

We also use it in the `help` menu to show the update status like this:
Screenshot 2024-09-23 at 11 30 19 AM

So I believe it makes more sense to unify the app update workflow to eliminate any confusion to the users. Wdyt? 

One slight disadvantage I can see with this approach: I'm not 100% certain if the native app update works well on Windows, as we don't have the app binary signed. In that case, the CTA in the notification might not do anything useful. But saying this only because I have personally not validated the native app update on Windows, it could be working fine as well. 🤷🏼‍♂️

violetadev:
@manojVivek 100%! I was thinking of trying to do this on a different pull request but seeing we have something already I’ll add it here. I have windows and mac, so I can try it out in both :D

violetadev:
@manojVivek I'm going to open a second PR with the update notification, so this PR doesn't go stale and it's not so big to review. Do you approve this one?

manojVivek:
> @manojVivek I'm going to open a second PR with the update notification, so this PR doesn't go stale and it's not so big to review. I removed the notification with the download link to the website. Now it's just the notification system. Do you approve?

Yes, sounds good! 

You can stack the PR with those changes on top of this PR and merge it altogether once everything is ready, whichever works for you.

manojVivek (approved):
Lgtm!

PR #1318, opened 2024-10-03
Suggest Chrome Extension Installation Popup

### LINKED ISSUE #1292: Suggest the installation of a Chrome extension if the user is using Chrome but doesn’t have the Responsively App installed.
# 🚀 Feature Request

### 📝 Description

Suggest the installation of a Chrome extension if the user is using Chrome but doesn’t have the Responsively App installed.

You could detect if the Responsively App extension is installed and offer an easy way to install it if not.

ashley44511:
# ✨ Pull Request

### 📓 Referenced Issue

resolves #1292 

### ℹ️ About the PR

A popup has been added to ask the user to install the "Responsively Helper" Chrome extension upon starting the Responsively App on desktop. The "Download Extension" button takes the user to the Chrome Web Store when clicked. The "I do not use Chrome" button marks the chromePopup.usesChrome variable false so that the popup will not appear again.

output

I tested using yarn to make sure the popup appears at the right time.

### DIFF
--- a/desktop-app/src/renderer/components/ChromePopup/index.tsx
+++ b/desktop-app/src/renderer/components/ChromePopup/index.tsx
@@ -1,0 +1,77 @@
+import { useEffect, useMemo, useState } from 'react';
+
+import { IPC_MAIN_CHANNELS } from 'common/constants';
+
+import Modal from '../Modal';
+import Icon from '../../assets/img/logo.png';
+import Button from '../Button';
+
+export const ChromePopup = () => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  useEffect(() => {
+    const usesChromeMs = window.electron.store.get('chromePopup.usesChrome');
+    if (usesChromeMs === undefined || usesChromeMs === true) {
+      setIsOpen(true);
+    }
+  }, []);
+
+  const onClose = () => {
+    setIsOpen(false);
+  };
+
+  return (
+    <Modal
+      title={
+        <div className="flex flex-col items-center gap-2 border-b border-slate-500 pb-2">
+          <div className="flex w-full justify-center">
+            <img src={Icon} alt="Logo" width={64} />
+          </div>
+          <div>Do you have our Chrome Extension?</div>
+        </div>
+      }
+      isOpen={isOpen}
+      onClose={onClose}
+    >
+      <div className="max-w-lg">
+        <p className="pb-4 text-center">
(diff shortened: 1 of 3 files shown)


desktop-app/src/renderer/components/ChromePopup/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+              onClick={() => {
+                window.electron.ipcRenderer.sendMessage(
+                  IPC_MAIN_CHANNELS.OPEN_EXTERNAL,
+                  {
+                    url: `https://chromewebstore.google.com/detail/responsively-helper/jhphiidjkooiaollfiknkokgodbaddcj`,
+                  }
+                );
+                window.electron.store.set('chromePopup.usesChrome', true);
+                onClose();
+              }}
```
violetadev, line 56: it is best practice to declare a new function outside of the onclick and then call it as in:
`onClick={handleDownload}`

If you declare a function inline in onClick={}, it will be recreated on every render. This can have performance implications

desktop-app/src/renderer/components/ChromePopup/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <div className="flex gap-1">
+            <Button
+              onClick={() => {
+                window.electron.ipcRenderer.sendMessage(
+                  IPC_MAIN_CHANNELS.OPEN_EXTERNAL,
+                  {
+                    url: `https://chromewebstore.google.com/detail/responsively-helper/jhphiidjkooiaollfiknkokgodbaddcj`,
+                  }
+                );
+                window.electron.store.set('chromePopup.usesChrome', true);
```
violetadev, line 54: usesChrome is a bit vague, I would name this something like `hasSeenChromeExtensionPopUp`

desktop-app/src/renderer/components/ChromePopup/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+            </Button>
+            <Button
+              onClick={() => {
+                window.electron.store.set('chromePopup.usesChrome', false);
+                onClose();
+              }}
+              isTextButton
+              isPrimary
+            >
+              I do not use Chrome
```
violetadev, line 70: I would change the text to "No, thank you", in order to cover more cases, like the person doesn't want it or they want to close the popup

PR #1340, opened 2024-12-10
fix: Fix restricted device size to enable phones landscape

### LINKED ISSUE #1210: When rotating to landscape, update the aspect ratio of the location on the main screen for some phone targets
# 🚀 Feature Request

### 📝 Description

When Rotating a iphone 12 in the main window to landscape mode, the width on screen stays the same and a the screen is not shown entirely.
 
When Rotating any device (or all devices) in the main window to landscape mode, the width on screen should be adjusted to reflect the new aspect ratios (possible as an opt-in option)

It works for ipad and macbook and also if I look at the iphone device indiviually the aspect ratio is correct

happymvp:
# ✨ Pull Request

### 📓 Referenced Issue

Fix #1210

### ℹ️ About the PR

Added checker for landscape mode before strict ("w-52") size of device. 

flex-shrink-0 is not needed because we have control over elements with overflow scroll.

But long-term I'd put all the layout settings (incl. Masonry Layout, Preview Layout) into one component. Now it's too easy to break the whole visual configuration with inappropriate css style

Screenshot 2024-12-10 at 15 45 20
Screenshot 2024-12-10 at 15 45 26
Screenshot 2024-12-10 at 15 45 33
Screenshot 2024-12-10 at 15 45 35

### DIFF
--- a/desktop-app/src/renderer/components/Previewer/Device/index.tsx
+++ b/desktop-app/src/renderer/components/Previewer/Device/index.tsx
@@ -72,7 +72,7 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
   const [singleRotated, setSingleRotated] = useState<boolean>(false);
   const [loading, setLoading] = useState<boolean>(false);
   const [error, setError] = useState<ErrorState | null>(null);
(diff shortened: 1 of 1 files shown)


violetadev:
awesome work, thank you!

Issue #1395, opened 2025-07-23
URL search bar broken, Unable to navigate


KierenFoenander:
# 🐞 bug report

### ✍️ Description

when using the application adn trying to go to a new page I get stuck in an endless loop of not being able to navigate to any other site after the first initial new address I put in, see gif below I will always be redirected to the page I am already on
![Image](https://github.com/user-attachments/assets/ff60c16b-01b9-41e2-83b8-bf2ed9d4613e)

### 🕵🏼‍♂️ Is this a regression?

Would assume so however unsure as this is the first time I am using 

### 🔬 Minimal Reproduction

1. off a fresh install open the app and put in your initial url you wish to go to.
2. click around as if you were trying to use the app as normal
3. attempt to use the url search bar again to go to a different address
4. get caught in loop of not being able to navigate

### 🌍 Your Environment

ctrl + f1 is not doing anything here  sorry. I am on latest version on windows

pratapmajge:
Hi! I would like to work on this issue. Could you please assign it to me?
If there are any specific guidelines or details I should follow, please let me know.

Kieren-Foenander:
> Hi! I would like to work on this issue. Could you please assign it to me? If there are any specific guidelines or details I should follow, please let me know.

Hey mate, sorry was already working on it, please check out the pr if you like though?

pratapmajge:
ok, go ahead mate 👍

PR #1396, opened 2025-07-24
Fix/url naviagtion not working

Kieren-Foenander:
# ✨ Pull Request

### 📓 Referenced Issue

Resolves: #1395 

### ℹ️ About the PR

fixes a race condition between addressBar navigation and didNavigateHandler overwriting the Redux store. Ensures addressbar navigation is dispatched properly and then cleaned up to allow for normal navigation.


manojVivek (approved):
Thank you for fixing this painstaking bug!

manojVivek:
@all-contributors Please add @Kieren-Foenander for code.

PR #1417, opened 2025-09-09
fix(electron): preserve window bounds on reload

rajanarahul93:
# ✨ Pull Request

### 📓 Referenced Issue

Fixes: #999

### ℹ️ About the PR

This PR resolves an issue on Windows where the application window would incorrectly maximize after a reload if it was snapped to the side of the screen.

The fix introduces a managed reload process:
1.  The "Reload" menu action in `menu/view.ts` no longer calls `reload()` directly. Instead, it sends an IPC message (`window-reload`) to the main process.
2.  A new listener in `main.ts` catches this message. It saves the window's current bounds (size and position), executes the reload, and then restores the saved bounds once the content has finished loading (on the `did-finish-load` event).

This ensures the window's state is preserved across reloads, providing a consistent user experience.


rajanarahul93:
Hi @manojVivek, this PR fixes the issue where the window would maximize after a reload on Windows. The reload action now preserves the window's bounds by saving its size and position before reload and restoring them once the content finishes loading. This ensures a consistent window state even when snapped. Happy to make any refinements if needed!

PR #1459, opened 2026-01-11
refactor: remove console logging statements for production build

Ayushnator:
- Remove all console.log() calls from production code to prevent sensitive information leakage
- Replace generic console.log() error handlers with proper console.error() for critical errors
- Update error handling in preload scripts to use external logging services instead
- Clean up developer debug statements from app initialization and migration logic
- Improve security and reduce console output noise in production application

# ✨ Pull Request


manojVivek:
These are some logs we would like to keep even in the production builds. So will close this one.

PR #1463, opened 2026-01-27
feat: Add trackpad pinch gesture support for zoom in preview area

### LINKED ISSUE #1049: Zoom in/out of devices on trackpad pinch gesture in the preview area

drewsephski:
- Implement touch event handlers for pinch-to-zoom functionality
- Add distance calculation between touch points for gesture detection
- Trigger existing zoomIn/zoomOut actions based on pinch scale factor
- Prevent default touch behavior during pinch gestures
- Fixes #1049

### DIFF
--- a/desktop-app/package-lock.json
+++ b/desktop-app/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "Responsively-App",
-  "version": "1.15.0",
+  "version": "1.17.1",
   "lockfileVersion": 3,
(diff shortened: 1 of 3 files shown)


manojVivek:
@drewsephski I just gave it a try and it doesn't seem to work for me on MacOS. Which OS did you try this on? 

Looks more like pinch events are delivered as wheel events on Mac.

Issue #1388, opened 2025-07-11
Spotify iframe - instant redirect


Kaienka:
Hi! First, thank you for the Responsively, this is great app, good job! It helps me a lot.
 
Unfortunately, I have a problem similar to this one: [App automatically navigates to iframe URL if lazy loaded](https://github.com/responsively-org/responsively-app/issues/164)

# 🐞 bug report

### ✍️ Description

I have a sub-page with embedded Spotify playlists iframes. When I open the page in the Resposively App, it is redirected instantly to one of the playlist, randomly chosen.

### 🕵🏼‍♂️ Is this a regression?

I don't know, I'm a new user.

### 🔬 Minimal Reproduction

1. Have embedded Spotify playlist on the website
2. Go to the page with iframes

Expected: 
It stays on the sub-page and embedded playlists are visible.

Current behaviour:
User is redirected immediately to the playlist.

### 🌍 Your Environment

App is written with Next.js, published on Vercel, iframes comes directly from Spotify share playlist option. Happens both on vercel and localhost development version. 
This is my app, feel free to check [Home page](https://kaishouri.vercel.app/), the playlists are at /music ("playlisty" menu position). 

Usage of iframe in my code:
<code>
  <iframe
        src={src}
        className={`rounded-xl w-full h-[22rem] transition-opacity duration-500 ${
          loaded ? "opacity-100" : "opacity-0 pointer-events-none"
        }`}
        frameBorder="0"
        allowFullScreen={false}
        allow="autoplay; clipboard-write; encrypted-media; picture-in-picture"
        loading="lazy"
      />
</code>

### 🔥 Exception or Error or Screenshot
No error occurred. 
[Recorded behaviour - Streamable](https://streamable.com/nrdx4b)

gjames50:
I'm currently experiencing the exact same behavior trying to embed famewall testimonials.

Sample Code:

```
<div class="famewall-embed" data-src="testimonial" data-format="grid" style="width:100%;"></div>
<script type="text/javascript" src="https://embed.famewall.io/frame.js" defer></script>
```

When the code is on the webpage I'm trying to view in the Responsively App, the page loads, then immediately redirects to: https://embed.famewall.io/wall/testimonial?autoplay=0

For now, I have to remove the code from the site to use the Responsively App.

Responsively App is on a Windows 11 Pro 24H2 computer. 

Hopefully, someone can sort this out. Thank you.

raestanton:
[...]

PR #1465, opened 2026-02-09
fix(webview): ignore iframe did-navigate-in-page events to prevent un…

bhnprksh222:
Hi @manojVivek, I’ve worked on this issue and pushed a fix.

**Issue:**
https://github.com/responsively-org/responsively-app/issues/1388

**Root cause:**
`did-navigate-in-page` events triggered by iframe loads (like Spotify embeds) were being treated as top-level navigations. This updated the address bar to the iframe `src` and caused Responsively to redirect away from the parent page.

**Fix:**
Added a guard so only **main-frame** `did-navigate-in-page` events update the address and history. Iframe navigations are now ignored, preventing unwanted redirects.

**Before (issue):**
Image

Image

**After (fixed):**

Image

Image

Tested with the reported Spotify embed page and also verified normal SPA navigation, back/forward, and reload flows.
Let me know if any changes are needed.

### DIFF
--- a/desktop-app/src/renderer/components/Previewer/Device/index.tsx
+++ b/desktop-app/src/renderer/components/Previewer/Device/index.tsx
@@ -290,7 +290,11 @@ const Device = ({ isPrimary, device, setIndividualDevice }: Props) => {
     const webview = ref.current as Electron.WebviewTag;
     const handlerRemovers: (() => void)[] = [];
 
-    const didNavigateHandler = (e: Electron.DidNavigateEvent) => {
+    const didNavigateHandler = (
+      e: Electron.DidNavigateEvent | Electron.DidNavigateInPageEvent
+    ) => {
+      // Only DidNavigateInPageEvent has isMainFrame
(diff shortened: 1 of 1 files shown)


bhnprksh222:
Tested with multiple iframes.

ScreenRecording2026-02-08at22 51 15-ezgif com-video-to-gif-converter

manojVivek (approved):
Looking great! 🚀

bhnprksh222:
Hi @manojVivek, thanks for the review and approval! Since the fix looks good on your end, should we go ahead and close this issue?

Issue: https://github.com/responsively-org/responsively-app/issues/1388

manojVivek:
Ah, I thought it was linked to the PR. Closed it now!

dandry:
I am still facing this issue, version details below:
App Version: 1.18.0
Electron Version: 27.0.4
Chrome Version: 118.0.5993.129
Node.js Version: 18.17.1
V8 Version: 11.8.172.17-electron.0
OS: Darwin arm64 25.5.0

PR #1481, opened 2026-03-25
feat: add javascript toggle controls for device previews

### LINKED ISSUE #490: [Feature] Turn off Javascript
Add buttons (per devices and for all) that toggle javascript in the device
This should be in the ` ⋮ ` button (_kebab menu_?) since it is not a usual action

**Hints:**
`<webview src="https://github.com" webpreferences="javascript=no"></webview>`
[`webpreferences`](https://www.electronjs.org/docs/api/webview-tag#webpreferences) is a comma separated list of strings, the full list of options can be found [here](https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions)

Dailin521:
## Summary
Add JavaScript toggle controls for device previews without remounting the React `Device` tree.

## Root cause
For #490, the `webview` must be recreated for JavaScript mode changes because `webpreferences` only take effect when the `webview` is created. The earlier implementation rebuilt the whole preview device subtree, which left stale listeners and stale webview references in follow-up flows.

## Changes
- keep each `Device` mounted and remount only the target `webview` when JavaScript is toggled
- refresh toolbar webview references and scope listener cleanup to the current webview instance
- keep screenshot actions disabled while JavaScript is disabled for a preview
- add a regression test to ensure toggling JavaScript does not remount `Device`

## Verification
- `npm run typecheck`
- `npx eslint src/renderer/components/Previewer/Device/index.tsx src/renderer/components/Previewer/index.tsx src/renderer/components/Previewer/index.test.tsx`
- `npm run test -- src/renderer/store/features/javascript/index.test.ts src/renderer/components/ToolBar/Menu/Flyout/JavaScriptControls/index.test.tsx src/renderer/components/Previewer/Device/Toolbar.test.tsx src/renderer/components/Previewer/index.test.tsx`
- `npm run test`
- manual Electron verification on 2026-03-25:
  - per-device JavaScript toggle remounts only the target `webview`
  - all-previews JavaScript toggle remounts all `webview`s
  - `Device` roots remain mounted in both flows
  - screenshot actions are disabled while JavaScript is off

Closes #490

### DIFF
--- a/desktop-app/e2e/tests/device-toolbar.spec.ts
+++ b/desktop-app/e2e/tests/device-toolbar.spec.ts
@@ -1,6 +1,8 @@
 import {ElectronApplication, Page} from '@playwright/test';
 import {test, expect} from '../fixtures/electron-app';
 
+const WEBVIEW_EXECUTION_FAILED = '__WEBVIEW_EXECUTION_FAILED__';
+
 const execInWebview = async (
   electronApp: ElectronApplication,
   wcId: number,
@@ -10,9 +12,13 @@ const execInWebview = async (
     async ({webContents}, {id, js}: {id: number; js: string}) => {
       const wc = webContents.fromId(id);
       if (!wc) {
-        throw new Error(`No webContents with id ${id}`);
+        return WEBVIEW_EXECUTION_FAILED;
       }
-      return wc.executeJavaScript(js);
+      try {
+        return await wc.executeJavaScript(js);
+      } catch (_error) {
+        return WEBVIEW_EXECUTION_FAILED;
+      }
     },
     {id: wcId, js: script}
   );
@@ -198,7 +204,7 @@ test.describe('Device Toolbar', () => {
           ),
         {timeout: 10_000}
       )
-      .toBe('undefined');
+      .toBe(WEBVIEW_EXECUTION_FAILED);
     await expect
       .poll(
         () =>
--- a/desktop-app/e2e/tests/device-toolbar.spec.ts
+++ b/desktop-app/e2e/tests/device-toolbar.spec.ts
@@ -155,16 +155,23 @@ test.describe('Device Toolbar', () => {
       )
       .toBe('number');
 
-    const quickScreenshotBtn = app.page.locator('button[title="Quick Screenshot"]').first();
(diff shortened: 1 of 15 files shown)


Dailin521:
Added the local runtime verification screenshots here for reviewer convenience.

These images are hosted in the dedicated public artifacts repository:
- https://github.com/Dailin521/oss-pr-artifacts

Per-device JavaScript toggle:
Real Electron verification passed. Disabled JavaScript on a single preview and confirmed only the targeted preview remounted.

Per-device JavaScript disabled verification

All previews JavaScript toggle:
Real Electron verification passed. Disabled JavaScript for all previews and confirmed the global control reached the expected disabled state.

All previews JavaScript disabled verification

Dailin521:
Quick follow-up here. CI is green and the current scope is still intentionally narrow, limited to the existing device preview JavaScript toggle paths. If this direction looks acceptable, I’m happy to make any small adjustments needed for merge.

manojVivek:
@Dailin521 Sorry for the delay in reviewing this, can you please check and make sure we have e2e test cases to cover this feature here: https://github.com/responsively-org/responsively-app/tree/main/desktop-app/e2e/tests?

Dailin521:
Added the requested E2E coverage on the updated head.

This now covers both paths under `desktop-app/e2e/tests`:
- the global `JavaScript (All Previews)` toggle in the menu flyout
- the per-preview `Disable/Enable JavaScript` action from the device toolbar overflow menu

The new tests also assert the related screenshot-disabled behavior and the preview remount behavior after toggling JavaScript.

Fresh `Analyze` and `test (macos-latest)` runs are already queued on this push.

Dailin521:
Pushed a follow-up fix for the first CI round. This only tightens the new E2E selectors so the test scopes to the first device toolbar more reliably and keeps the screenshot-all assertion stable when the disabled tooltip text changes. Fresh checks are running now.

Dailin521:
Pushed one more follow-up for the second CI round. The remaining failures were not selector-related anymore: on javascript=no webviews, Electron can reject executeJavaScript calls instead of returning a page-level undefined, so the new E2E assertions now treat that disabled-preview path as an expected execution failure signal and keep the enabled-preview checks intact. Fresh checks are running again.

PR #1499, opened 2026-05-07
Fix individual device rotation button

### LINKED ISSUE #1492: The individual rotation button doesn’t work; only the general rotation button works. v1.18.0
This image shows the buttons for an iPad, where the individual rotation option does not work.
Image

This image shows the general buttons next to the address bar; in this case, the rotate button does work.
Image

This image shows the iPad buttons after using the general rotate button. In this case, the rotate button becomes enabled, but it still does not work.
Image

Overall, it seems that the individual rotation buttons for mobile devices do not work; only the general rotation button works. I am using version 1.18.0; in the previous version, they worked correctly.

BboyGT:
## Summary
- enables the per-device rotate button for mobile-capable devices even before they are rotated
- keeps non-mobile devices disabled
- removes the redundant toolbar prop that used the current rotated state as the availability check

Fixes #1492.

## Testing
- Ran git diff --check
- Not run locally: project dependencies are not installed in this checkout

### DIFF
--- a/desktop-app/src/renderer/components/Previewer/Device/Toolbar.tsx
+++ b/desktop-app/src/renderer/components/Previewer/Device/Toolbar.tsx
@@ -20,7 +20,6 @@ interface Props {
   onRotate: (state: boolean) => void;
   onIndividualLayoutHandler: (device: Device) => void;
   isIndividualLayout: boolean;
-  isDeviceRotationEnabled: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


BboyGT:
Happy to adjust this if you prefer the toolbar to receive a separate canRotateDevice prop instead. I’m also happy to keep helping with nearby preview/device UI issues.

Issue #377, opened 2020-08-04
Add an option to do text search in the rendered devices


manojVivek:
1. `Cmd/Ctrl` + `F` should open the search box like chrome.
<img width="359" alt="Screenshot 2020-08-04 at 11 05 24 PM" src="https://user-images.githubusercontent.com/1283424/89326141-79d36100-d6a7-11ea-82f9-3fe0f37397d7.png">

2. Any user interaction can invoke the electron API to highlight the search text: https://www.electronjs.org/docs/api/web-contents#contentsfindinpagetext-options

romanakash:
Can I work on this?

esprush:
@romanakash I have started on this. my bad did not update the status.

PR #1532, opened 2026-08-02
Adds Common Find Text to search across all the devices

esprush:
# ✨ Pull Request

### 📓 Referenced Issue
#377 

### ℹ️ About the PR

This adds ability to search the contents of the Webpages loaded across all the devices.

#### Note:
**Freeze Issue in Previous implementation**

This tries to overcome the app freeze limitation in the previous implementation #400.
Previous implementation used `webview.findInPage` which internally had issues causing the app to freeze.
The newer version of uses `WebContent.findInPage`, which is different from the previous problematic option.

**FullScreen Conflict** 
In some cases when focus is inside a webview, `Cmd + F` opens webview FullScreen instead of Find Text

image


desktop-app/.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged
+cd desktop-app && npx lint-staged
```
esprush, line 4: commit was failing without this 🤔, @manojVivek

